PR #155, opened 2017-01-30
add python-numpy run dependenncy

MatthijsBurgh:


reinzor:
Build is failing apparently ... We need to fix this first ..

LoyVanBeek:
The .travis.yaml of this has always been borked. I 'fixed' it multiple times but because it depends on everything, it breaks when anything breaks. 
This new dependency is not making that difference.

reinzor:
I agree but then just work from the ground up to fix everything and not just accept this :)

-Rein

LoyVanBeek:
It's in the works but takes a lot of effort and waiting time and then it breaks for no good reason.

PR #160, opened 2017-02-11
wip: Refactor/cleanup and merge

jlunenburg:


challenge_manipulation/src/manipulation.py
```diff
@@ -45,7 +45,7 @@
 from empty_shelf_designator import EmptyShelfDesignator
 
 # PDF writer
-import pdf
+# import pdf
```
jlunenburg: Don't we need the pdf writer?

challenge_manipulation/src/manipulation.py
```diff
@@ -682,7 +664,7 @@ def export_to_pdf(userdata):
             entities = [ e[0] for e in DETECTED_OBJECTS_WITH_PROBS ]
 
             # Export images (Only best MAX_NUM_ENTITIES_IN_PDF)
-            pdf.entities_to_pdf(robot.ed, entities[:MAX_NUM_ENTITIES_IN_PDF], "tech_united_manipulation_challenge")
+            # pdf.entities_to_pdf(robot.ed, entities[:MAX_NUM_ENTITIES_IN_PDF], "tech_united_manipulation_challenge")
```
jlunenburg: Why not?

challenge_manipulation/src/manipulation.py
LoyVanBeek: Ouch, reviews already paying off. 
I outcommented that stuff because it was giving me errors while testing. Something with xhtml. 

The offending commit should be reverted.

challenge_manipulation/src/manipulation.py
LoyVanBeek: Ouch, reviews already paying off. 
I outcommented that stuff because it was giving me errors while testing. Something with xhtml.

challenge_manipulation/src/manipulation.py
LoyVanBeek: It even says so in the commit message: 
https://github.com/tue-robotics/tue_robocup/commit/ab3cb997b41bb4613a417532e54d29ea72330417

PR #162, opened 2017-02-14
Refactor/kdl datatypes

LoyVanBeek:
Addresses https://github.com/tue-robotics/tue_robocup/issues/156 

This allows to perform calculations with the geometric datatypes directly and allows to reduce geometric calc. code all around.

Issue #156, opened 2017-01-31
Use KDL Datatypes instead of geometry_msgs


LoyVanBeek:

LoyVanBeek:
# .base.get_location()
challenge_speech_recognition/src/indirect_speech_recognition.py
- [ ] 56:                                                        center_point=self.robot.base.get_location().p)
- [ ] 67:        current = self.robot.base.get_location()

challenge_final/src/challenge_final/handover_sergio.py
- [ ] 110:        # base_pose_map_msg = sergio.base.get_location()  #

robot_smach_states/src/robot_smach_states/manipulation/grasp_point_determination.py
- [ ] 50:        robot_pose = self._robot.base.get_location()

robot_smach_states/src/robot_smach_states/console-nl.py
- [ ] 148:    robot_pos = robot.base.get_location().pose.p

robot_smach_states/src/robot_smach_states/navigation/follow_operator.py
- [ ] 88:        current_pose_stamped = self._robot.base.get_location()
- [ ] 194:        robot_position = self._robot.base.get_location().pose.position
- [ ] 261:            robot_position = self._robot.base.get_location().pose.position
- [ ] 270:            robot_position = self._robot.base.get_location().pose.position
- [ ] 285:                    robot_position = self._robot.base.get_location().pose.position
- [ ] 349:        robot_position = self._robot.base.get_location().pose.position
- [ ] 493:        robot_position = self._robot.base.get_location().pose.position
- [ ] 587:            if (self._robot.base.get_location().header.stamp - self._time_started).to_sec() > self._start_timeout:

robot_smach_states/src/robot_smach_states/navigation/navigation.py
- [ ] 28:        self.robot.base.analyzer.start_measurement(self.robot.base.get_location())
- [ ] 40:        self.robot.base.analyzer.stop_measurement(self.robot.base.get_location(), self.result)

robot_smach_states/src/robot_smach_states/navigation/navigate_to_explore.py
- [ ] 37:        current_pose = self.robot.base.get_location()

robot_smach_states/src/robot_smach_states/utility.py
- [ ] 39:        self.robot.base.get_location()

challenge_following_and_guiding/src/challenge_following_and_guiding.py
- [ ] 27:        self._robot.ed.update_entity(id="robocup_arena", posestamped=self._robot.base.get_location(), type="waypoint")

challenge_open/src/find_person.py
- [ ] 116:            bp = self._robot.base.get_location()

challenge_restaurant/src/automatic_side_detection.py
[...]

PR #163, opened 2017-02-14
Refactor/kdl datatypes

LoyVanBeek:
Addresses #156

This allows to perform calculations with the geometric datatypes directly and allows to reduce geometric calc. code all around.

As opposed to https://github.com/tue-robotics/tue_robocup/pull/162, this PR does merge into the correct branch: refactor/cleanup_and_merge

### DIFF
--- a/robot_skills/src/robot_skills/util/entity.py
+++ b/robot_skills/src/robot_skills/util/entity.py
@@ -2,6 +2,7 @@
 import yaml
 
 # ROS
+import rospy
 import PyKDL as kdl
 
 from robot_skills.util.kdl_conversions import poseMsgToKdlFrame, kdlFrameToPoseMsg
@@ -112,6 +113,7 @@ def from_entity_info(e):
 
     # The data is a string but can be parsed as yaml, which then represent is a much more usable data structure
     volumes = volumes_from_entity_info_data(yaml.load(e.data))
+    rospy.logdebug("Entity(id={id}) has volumes {vols} ".format(id=identifier, vols=volumes.keys()))
 
     super_types = e.types
 
--- a/robot_skills/src/robot_skills/util/entity.py
+++ b/robot_skills/src/robot_skills/util/entity.py
@@ -4,7 +4,7 @@ import yaml
 # ROS
 import PyKDL as kdl
 
-from robot_skills.util.kdl_conversions import pose_msg_to_kdl_frame, kdl_frame_to_pose_msg
+from robot_skills.util.kdl_conversions import poseMsgToKdlFrame, kdlFrameToPoseMsg
 from robot_skills.util.volume import volumes_from_entity_info_data
 from robot_skills.util.shape import shape_from_entity_info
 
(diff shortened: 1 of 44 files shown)


robot_skills/src/robot_skills/util/entity.py
```diff
@@ -88,12 +89,12 @@ def is_a(self, super_type):
     @property
     def pose(self):
         """ Returns the pose (at this point, as a geometry_msgs.Pose, in the future as a kdl frame """
```
LoyVanBeek: I'll grep for this at some point, because there should not be a need for this any more.

robot_skills/src/robot_skills/util/kdl_conversions.py
```diff
@@ -1,59 +1,245 @@
...
+    def __init__(self, frame, frame_id, stamp=None):
+        assert isinstance(frame, kdl.Frame)
+        self.frame = frame
+        self.frame_id = frame_id
 
-    :param quat: geometry_msgs.msg.Quaternion
-    :return: kdl Rotation
-    """
-    return kdl.Rotation.Quaternion(quat.x, quat.y, quat.z, quat.w)
+    def __repr__(self):
```
LoyVanBeek: I used the first thing that worked. A better one-line representation is welcome. 
Preferable, some representation that can be copy-pasted into a console to construct the same value.

robot_skills/src/robot_skills/util/kdl_conversions.py
```diff
@@ -1,59 +1,245 @@
...
+        return "{frame} @ {fid}".format(frame=self.frame, fid=self.frame_id)
 
+    def projectToFrame(self, frame_id, tf_listener):
```
LoyVanBeek: This is a habit from work @Ultimaker. the under_score is nicer to be honest. 
At least this is easily grep-able and thus replaceable :-/

robot_skills/src/robot_skills/util/nav_analyzer.py
```diff
@@ -142,7 +142,7 @@ def stop_measurement(self, endpose, result):
 
         ''' Log endpose '''
         endposeitem = ET.SubElement(self.logitem, "endpose")
-        self.poseStampedToSubElement(endpose, endposeitem)
```
LoyVanBeek: It's only used in count_{reset,clear,plan}, which are themselves never used in that file so the whole lot can be removed.

robot_skills/src/robot_skills/world_model_ed.py
```diff
@@ -26,13 +26,17 @@
 import yaml
 
 from .classification_result import ClassificationResult
-
 from robot_skills.util.entity import from_entity_info
 
+def _create_service_proxy(service_name, service_type, wait_service):
```
LoyVanBeek: I cannot remember making this at all, I think I pulled in from merging master.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -210,10 +195,7 @@ def execute(self, userdata):
             roll = 0.3
         else:
             roll = -0.3
```
LoyVanBeek: Nope, it isn't so this must definitely be fixed

PR #166, opened 2017-02-22
Feature/set init pose on challenge startup

reinzor:
I would like to combine some of the varying middleware components with
the challenge since they are challenge specific. By doing it this way,
we launch the challenge with one launch file together with the
dependencies. Setting the initial pose at challenge start is required
since we do not have time now to do this with rviz.


MatthijsBurgh:
Initial pose is always set by start_challenge_robust. Or not?

reinzor:
Apparently not lol =D

-Rein

PR #181, opened 2017-03-10
Feature/arm tests

LoyVanBeek:


LoyVanBeek:
The test code is tested on Amigo and it works.
There however is an issue with the 'photo' pose (also in simulation):

The photo pose is [-1.4, -0.4, 0.0, 0.7, 0.0, 0.4, 0.1] but somehow some node complains that "Joint 'wrist_yaw_joint_left' goes out of limits in point 0 (min = -0.61, max = 0.61, requested goal = 30)" 
Which is odd because there is no mention of a 30 rad angle anywhere in the pose.

LoyVanBeek:
Actually, there was a 30 in another goal, not related to the photo pose. That just took too long.
The 30 is fixed in https://github.com/tue-robotics/amigo_description/pull/8

robot_skills/test/test_arms.py
```diff
@@ -0,0 +1,81 @@
...
+
+import robot_skills.util.kdl_conversions as kdl_conversions
+
+if len(sys.argv) < 2:
+    print "Please specify a robot name"
+    sys.exit()
+
+robot_name = sys.argv[1]
+
+rospy.init_node("audio_test")
```
jlunenburg: Bijzondere naam ;)

robot_skills/test/test_arms.py
```diff
@@ -0,0 +1,81 @@
...
+if robot_name == "amigo":
+    from robot_skills.amigo import Amigo
```
jlunenburg: This code is duplicated a couple of times throughout tue_robocup. Maybe refactor/re-use?

robot_skills/test/test_arms.py
```diff
@@ -0,0 +1,81 @@
...
+    if arm == robot.rightArm:
+        goal1 = kdl_conversions.kdlFrameStampedFromXYZRPY(0.342, -0.125, 0.748, 0, 0, 0, "/" + robot.robot_name + "/base_link")
+
+    robot.speech.speak("Moving {} arm to dummy goal pose".format(side))
+    if not arm.send_goal(goal1):
+        robot.speech.speak("{} arm could not reach dummy goal pose".format(side))
+        failed_actions += [goal1]
+    arm.wait_for_motion_done()
+
+    # TODO: Now check that the hand frame is within some tolerance to the desired goal.
```
jlunenburg: If this is available, please add it already, would indeed be usefull

PR #184, opened 2017-03-11
feature(RobotSkills) All parts inherit from RobotPart and all connect…

jlunenburg:
…ions are checked

### DIFF
--- a/robot_skills/src/robot_skills/ears.py
+++ b/robot_skills/src/robot_skills/ears.py
@@ -3,17 +3,18 @@ import rospy
 from dragonfly_speech_recognition.msg import Choice
 from dragonfly_speech_recognition.srv import GetSpeech, GetSpeechRequest
 
+from robot_part import RobotPart
 
-class Ears:
+
+class Ears(RobotPart):
     """
     Interface to amigo Ears.
     Works with dragonfly speech recognition (as of december 2014)
 
     Function listen explained on wiki: http://servicerobot.cstwiki.wtb.tue.nl/index.php?title=Using_the_dragonfly_speech_recognition
     """
-
-    def __init__(self, robot_name, pre_hook=None, post_hook=None):
-        self._get_speech_client_service = rospy.ServiceProxy("/%s/speech_client/get_speech"%robot_name, GetSpeech)
+    def __init__(self, robot_name, tf_listener, pre_hook=None, post_hook=None):
+        super(Ears, self).__init__(robot_name=robot_name, tf_listener=tf_listener)
         self._pre_hook = pre_hook
         self._post_hook = post_hook
 
--- a/robot_skills/src/robot_skills/head.py
+++ b/robot_skills/src/robot_skills/head.py
@@ -1,17 +1,18 @@
 #! /usr/bin/env python
 import rospy
-from actionlib import SimpleActionClient
 from geometry_msgs.msg import PointStamped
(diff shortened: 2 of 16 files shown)


robot_skills/src/robot_skills/ears.py
```diff
@@ -3,17 +3,18 @@
...
+class Ears(RobotPart):
     """
     Interface to amigo Ears.
     Works with dragonfly speech recognition (as of december 2014)
 
     Function listen explained on wiki: http://servicerobot.cstwiki.wtb.tue.nl/index.php?title=Using_the_dragonfly_speech_recognition
     """
-
-    def __init__(self, robot_name, pre_hook=None, post_hook=None):
-        self._get_speech_client_service = rospy.ServiceProxy("/%s/speech_client/get_speech"%robot_name, GetSpeech)
```
LoyVanBeek: No need for the _get_speech_client_service anymore?

robot_skills/src/robot_skills/head.py
```diff
@@ -34,39 +35,39 @@ def look_at_hand(self, side):
         Optionally, keep tracking can be disabled (keep_tracking=False)
         """
         if side == "left":
-            return self.look_at_point(msgs.PointStamped(0,0,0,frame_id="/"+self._robot_name+"/grippoint_left"))
+            return self.look_at_point(msgs.PointStamped(0,0,0,frame_id="/"+self.robot_name+"/grippoint_left"))
```
LoyVanBeek: How come these are still PointStampeds?

robot_skills/src/robot_skills/robot.py
```diff
@@ -32,38 +33,53 @@ class Robot(object):
...
         self.base_link_frame = "/"+self.robot_name+"/base_link"
 
-        #Grasp offsets
+        # Grasp offsets
         #TODO: Don't hardcode, load from parameter server to make robot independent.
         self.grasp_offset = geometry_msgs.msg.Point(0.5, 0.2, 0.0)
 
+        # Create attributes from dict
+        for k, v in self.parts.iteritems():
+            setattr(self, k, v)
```
LoyVanBeek: This is indeed much nicer than override __getattr__ to dynamically look up the parts: this gives tab completion.

But instantiating them the old way and then putting in a dictionary or list also works. 

You could also iterate over all members and then check if they are an instance of RobotPart with isinstance(part, RobotPart):
```python
for part in [part for part in self.__dict__.values() if isinstance(part, RobotPart):
    part.wait_for_connections(1.0)
```

robot_skills/src/robot_skills/ears.py
LoyVanBeek: Ah, not in Ears, obvisously

robot_skills/src/robot_skills/util/kdl_conversions.py
```diff
@@ -35,6 +35,23 @@ def extractVectorStamped(self):
         return VectorStamped(frame_id=self.frame_id, vector=self.frame.p)
 
 
+def frame_stamped(frame_id, x, y, z, roll=0, pitch=0, yaw=0):
```
LoyVanBeek: There already is a FrameStampedFromXYZRPY on the bottom of this file. Granted, this function is easier on the eyes. 

I really need to get rid of this CamelCase bug in my head

PR #198, opened 2017-03-15
Change timeout in Torso, because of changes in TrajectoryActionLib

MatthijsBurgh:


MatthijsBurgh:
To be sure, this should be merged unless frequency of rtt trajectory action lib is changed.

MatthijsBurgh:
@reinzor To be sure, this should not be merged unless the frequency of rtt trajectory action lib is changed. It can be changed, but you are introducing a unnecessary delay.

PR #218, opened 2017-04-08
Feature/base get location returns framestamped

### LINKED ISSUE #208: robot.base.get_location should return a FrameStamped
This stamp will be always be /map probably but explicit is better than implicit. It also reduces some code otherwhere. 

Suggested by @timassman

LoyVanBeek:
Fix #208

### DIFF
--- a/robot_smach_states/src/robot_smach_states/navigation/follow_operator.py
+++ b/robot_smach_states/src/robot_smach_states/navigation/follow_operator.py
@@ -88,21 +88,21 @@ class FollowOperator(smach.State):
         return False
 
     def _standing_still_for_x_seconds(self, timeout):
-        current_pose_stamped = self._robot.base.get_location().frame
+        current_pose = self._robot.base.get_location().frame
         now = rospy.Time.now()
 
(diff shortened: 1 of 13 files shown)


robot_smach_states/src/robot_smach_states/navigation/follow_operator.py
```diff
@@ -88,7 +88,7 @@ def _operator_standing_still_for_x_seconds(self, timeout):
         return False
 
     def _standing_still_for_x_seconds(self, timeout):
-        current_pose_stamped = self._robot.base.get_location()
+        current_pose_stamped = self._robot.base.get_location().frame
```
jlunenburg: This is called stamped, while you actually remove the stamp. Although it was 'wrong' before, I'd change it

LoyVanBeek:
Tested with r5cop demo, that works like a charm

LoyVanBeek:
@jlunenburg can you re-evaluate the PR? Review comment are fixed and code is tested with R5Cop demo

PR #238, opened 2017-04-18
Add --debug switch to enable the introspection server

Rayman:
This will allow users to kill challenges, cool huh?


LoyVanBeek:
Srsly, the introspection-server breaks ctrl-C?

PR #258, opened 2017-05-17
WIP: Refactor/entity pose as framestamped

LoyVanBeek:
This makes Entity.pose a FrameStamped so usage is much simpler without the need to add the stamp yourself. 

The benefit is most clear is cases like https://github.com/tue-robotics/action_server/commit/64d66858888f87eeec8a5513a3d4cea27c144655

Testing required:

- [ ] robot_skills tests, maybe more tests are needed that eg. integrate ED and head to look at an entity
- [ ] robot_smach_states via e.g. storing_groceries or help_me_carry


robot_smach_states/src/robot_smach_states/navigation/navigate_to_grasp.py
```diff
@@ -52,14 +52,14 @@ def generateConstraint(self):
...
-            y = pose.position.y
+            x = pose.frame.p.x()
+            y = pose.frame.p.y()
         except KeyError, ke:
             rospy.logerr("Could not determine pose: ".format(ke))
             return None
 
         try:
-            rz = entity.pose.orientation.z
+            rz, _, _ = entity.pose.frame.M.GetEulerZYX()
```
rokusottervanger: If there are also nonzero roll and pitch angles, this will generally not return the correct projected yaw angle as the other angles may also influence this. I guess this is out of the scope of this PR, as this problem already existed in the old version, but an issue should be filed for this.

LoyVanBeek: I'm not sure in which cases this happens, can you file that issue?

rokusottervanger: Check! https://github.com/tue-robotics/tue_robocup/issues/260

LoyVanBeek:
Wow, some of the checks actually passed for once :smile:

LoyVanBeek:
I did a rebase on top of rgo2017 to get the updates in knowledge etc in there as well

PR #274, opened 2017-05-20
open of RGO2017 working again

MatthijsBurgh:


challenge_open/src/challenge_open.py
```diff
@@ -44,26 +40,6 @@ def execute(self, userdata):
         return "done"
 
 
-class StartDetectingSkeletonsForever(smach.State):
```
LoyVanBeek: Would it make sense to put this in robot_smach_states, in a new section for gestures etc.

MatthijsBurgh: Yes, That is a good idea

robocup_knowledge/src/robocup_knowledge/environments/robotics_testlabs/challenge_open.py
```diff
@@ -1,23 +1,23 @@
-from itertools import groupby
 from robocup_knowledge import knowledge_loader
 common = knowledge_loader.load_knowledge("common")
 
-initial_pose = "initial_pose_door_B"  # initial pose
-starting_point = "eegpsr_starting_pose"  # Designated pose to wait for commands
-exit_waypoint = "exit_door_B1"
-ask_waypoint = "eegpsr_starting_pose"
+# initial_pose = "initial_pose_door_B"  # initial pose
```
Rayman: Just remove this if you're not using it

PR #276, opened 2017-05-22
Feature/check ebutton

LoyVanBeek:


robot_smach_states/src/robot_smach_states/check_ebutton.py
```diff
@@ -0,0 +1,36 @@
...
+
+        with self:
+            smach.StateMachine.add( "CHECK_EBUTTON",
+                                    CheckEButton(robot),
+                                    transitions={   "pressed"   :"SAY_PRESSED",
+                                                    "released"  :"succeeded"})
+
+            smach.StateMachine.add("SAY_PRESSED",
+                                    Say(robot, ["My Emergency button is still pressed", "EButton still pressed"], block=False),
+                                    transitions={   "spoken":"succeeded"})
```
MatthijsBurgh: I would go to a sleep of like 5 seconds and than check again. @jlunenburg What do you think is a robust solution?

LoyVanBeek: The robust solution is to not be a dumb-ass :-). There is a small delay already due to the 
`block=True` in the `Say`-state

Rayman:
How are we going to use this? Immediately after startup (utils/startup)?

LoyVanBeek:
That is indeed the idea. It is already done in StartChallengeRobust but not all challenges use this.

jlunenburg:
Add topics

LoyVanBeek:
The checking of all body parts is a separate issue: https://github.com/tue-robotics/tue_robocup/issues/283

robot_skills/src/robot_skills/util/robot_constructor.py
```diff
@@ -1,5 +1,10 @@
-def robot_constructor(robot_name):
+import sys
+
+def robot_constructor(robot_name=None):
     """Construct a robot by it's name. Choices are amigo, sergio, mockbot"""
+    if not robot_name:
+        robot_name = sys.argv[1]
```
Rayman: This file does nothing with argument parsing. Why add it here>

LoyVanBeek: This way, you can just call 

```python
robot = robot_constructor()
```

instead of the slightly more lines of code:
```python
import sys
robot_name = sys.argv[1]
robot = robot_constructor(robot_name)
```

But I agree the advantage is very small.

jlunenburg: +1 for this comment. It makes assumptions about the way this code is used and in general this seems not a sensible thing to do.

LoyVanBeek: OK, I'll remove this bit

LoyVanBeek: OK: https://github.com/tue-robotics/tue_robocup/pull/298

PR #284, opened 2017-05-23
Fix/gpsr knowledge

rokusottervanger:


robocup_knowledge/src/robocup_knowledge/environments/robotics_testlabs/challenge_gpsr.py
```diff
@@ -1,10 +1,258 @@
...
+    else:
+        sentence = grammar_parser.get_random_sentence("T")
+
+    print "Parsing sentence:\n\n{}\n\n".format(sentence)
+
+    result = grammar_parser.parse("T", sentence)
+
+    print "Result:\n\n{}".format(result)
 
-translations = { "bookcase" : "bocase" }
```
LoyVanBeek: Though it looks funny, I think this was added with a reason

PR #290, opened 2017-05-23
Feature/check estop for real robot

LoyVanBeek:


Timple:
Would it be hard to add the ebutton to the simulator?

This would potentially scale better in the future because otherwise we need to check the ROBOT_REAL variable everywhere.

LoyVanBeek:
Should be simple, just a publisher.

PR #292, opened 2017-05-23
fix(robotics_testlabs/common): names lower case for hmi to work

rokusottervanger:
This is apparently necessary for the hmi to work. Should actually be fixed in the hmi.


rokusottervanger:
@MatthijsBurgh , please only review the PR, leave merging to the person that opened the PR

PR #293, opened 2017-05-23
fix(robotics_testlabs/common): add room based manipulation location

rokusottervanger:
Based on a mapping of rooms to most probable manipulation locations. In the future, this will be replaced by searching for objects at all locations in the given room.


LoyVanBeek:
Does it make sense to already add an issue to implement the next step "searching for objects at all locations in the given room"?

PR #302, opened 2017-05-30
Feature/wait for click

Rayman:
Drive to the clicked_point as customer. Works in simulation

### DIFF
--- a/challenge_restaurant/src/challenge_restaurant/restaurant.py
+++ b/challenge_restaurant/src/challenge_restaurant/restaurant.py
@@ -10,7 +10,7 @@ import robot_smach_states as states
 # Challenge restaurant
 from store_waypoint import StoreWaypoint
 from take_orders import TakeOrder, ReciteOrders
-from wait_for_customer import WaitForCustomer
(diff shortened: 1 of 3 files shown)


LoyVanBeek:
Also works on real robot.
One additional thing is that in RViz, you can't just click somewhere to make a 3D point, you need something to intersect with. If there is no object yet seen, there is nothing to click on.

Perhaps adding a marker for ground plane is helpful.

challenge_restaurant/src/challenge_restaurant/restaurant.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, robot):
                                    transitions={'done': 'WAIT_FOR_CUSTOMER'})
 
             smach.StateMachine.add('WAIT_FOR_CUSTOMER',
-                                   WaitForCustomer(robot, caller_id),
```
LoyVanBeek: Can we instantly switch between these? It would be nice to be able to do both with no code changes.

LoyVanBeek:
We need to reimplement the whole waving stuff, so currently this is the only thing that works. Approving

PR #312, opened 2017-06-08
userdata=None

MatthijsBurgh:


MatthijsBurgh:
Sorry for the one wrong commit message, too quick with ctrl+c & ctrl+v :p

robot_smach_states/examples/random_nav2.py
```diff
@@ -189,7 +189,7 @@ def __init__(self, robot):
             @smach.cb_interface(outcomes=['target_determined', 'no_targets_available'],
                                 input_keys=[],
                                 output_keys=[])
-            def determine_target(userdata, designator):
+            def determine_target(userdata=None, designator):
```
LoyVanBeek: Python does not allow keyword arguments before other arguments AFAIK

Rayman:
Can you explain why you made this PR?

MatthijsBurgh:
So you can use smach.states outside a state machine

MaxBaeten:
You can just do:
state.execute({})

MatthijsBurgh:
@LoyVanBeek, what needs to be changed? We discussed it last Tuesday, but not thoroughly

LoyVanBeek:
There are 3 options:

- Do not do the ```userdata=None```, but leave just ```userdata```. This requires the user to have a knowledge of python
- also set ```designator=None``` but this might cause issues because None makes no sense. Perhaps a different default argument but I can't think of a sensible value
- Swap the order of arguments, but then you also have to check all callers to also change this order.

PR #313, opened 2017-06-08
Switch to docker CI

Timple:
Should work when https://github.com/tue-robotics/tue-env/pull/98 is fixed.

Last two commits here are tests:
https://travis-ci.org/tue-robotics/tue_robocup/builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
   - wget https://raw.githubusercontent.com/tue-robotics/tue-env/master/ci/install-package.sh
   - wget https://raw.githubusercontent.com/tue-robotics/tue-env/master/ci/build-package.sh
   # Determine packages modified in this commit
-  - PACKAGES=`git diff-tree --name-only $TRAVIS_COMMIT | xargs ls -dl 2>/dev/null |  grep "^d" | awk '{print $NF}' `
+  - PACKAGES=`git diff-tree --name-only $TRAVIS_COMMIT | xargs ls -dl 2>/dev/null |  grep "^d" | grep -v "\." | awk '{print $NF}' `
 
 script: 
   - 'for PACKAGE in $PACKAGES;
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,40 +1,19 @@
-sudo: required
-dist: trusty
-language: generic
-env:
-  - ROS_DISTRO=indigo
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,40 +1,19 @@
...
-  # install travis' package dependencies
-  - cd "$TUE_SYSTEM_DIR/src"
-  - rm *
-  - tue-get update ros-$REPOSITORY_NAME
-  - ls -l
-
-  # go to the catkin workspace
-  - cd "$TUE_SYSTEM_DIR"
+before_install:
+  - wget https://raw.githubusercontent.com/tue-robotics/tue-env/master/ci/install-package.sh
```
LoyVanBeek: Can we make this build trigger as well when tue-env is updated? 
Via http://eng.rightscale.com/2015/04/27/dependent-builds-in-travis.html or https://github.com/mernst/plume-lib/blob/master/bin/trigger-travis.sh

Timple: We can! Feel free to put in a new issue.

Especially the upstream is quite hard to parse in our current setup. Downstream is doable (and actually already performed by tue-get).

LoyVanBeek: Upstream could be done with ```rospack depends-on``` for ROS packages.

.travis.yml
```diff
@@ -1,40 +1,19 @@
...
+  - wget https://raw.githubusercontent.com/tue-robotics/tue-env/master/ci/build-package.sh
+  # Determine packages modified in this commit
+  - PACKAGES=`git diff-tree --name-only $TRAVIS_COMMIT | xargs ls -dl 2>/dev/null |  grep "^d" | awk '{print $NF}' `
 
-script:
-  - catkin build
-  - catkin run_tests # and run them
```
LoyVanBeek: Run ```catkin_run_tests``` in the docker image

.travis.yml
LoyVanBeek: Issue filed: #314.

.travis.yml
LoyVanBeek: Something like 
```docker run -it -d tuerobotics/tue-env catkin run_tests```

Timple: That is a good point!

https://github.com/tue-robotics/tue-env/pull/99

PR #316, opened 2017-06-09
Feature/set a table

SamAlexandrov:
I accidentally pushed first to this branch, then to the feature/demo.
Please delete this or merge it, I'm not sure what should be done :)


LoyVanBeek:
Can the branch also be deleted? Additions to set-a-table will be separate branches, right?

PR #326, opened 2017-06-15
Open of RGO 2017 working in robotics_testlabs

MatthijsBurgh:


challenge_open/src/challenge_open.py
```diff
@@ -44,26 +40,6 @@ def execute(self, userdata):
         return "done"
 
 
-class StartDetectingSkeletonsForever(smach.State):
```
LoyVanBeek: Won't we need this again eventually? Maybe in a different place though.

MatthijsBurgh: Where do you want to put it?

LoyVanBeek: If so, then in robot_smach_states, but for now here is fine. 

And actually it's kind of a weird thing, cuz' maybe this should be in a separate service or whatever to just get the latest upon request. 
For now, just keep that code in, I think.

PR #332, opened 2017-06-27
added a check to compare the objects in common knowledge and labels as Janno requested

Jip238:
…s Janno requested


test_tools/data_collection/analyse-annotations
```diff
@@ -70,9 +70,28 @@ if __name__ == "__main__":
...
+
+    print "objects that are present in objects_set but not in tensorflow_set\n"
+    for o in objects_set.difference(tensorflow_set):
+        print o
+    print "\n"
+
+    print "objects that are present in tensorflow_set but not in objects_set\n"
+    for t in tensorflow_set.difference(objects_set):
+        print t
+    print "\n"
```
LoyVanBeek:
```python
print
``` 

also prints a newline

MatthijsBurgh: So remove all the "\"??

LoyVanBeek:
```python
print "\n"
```
can be replaced with a simple 

```python
print
```

LoyVanBeek:
I approved but then I noticed the extension to the file is missing. No biggie, but I'd add it for consistency

MatthijsBurgh:
We have files with and without extension. Do we have any policy for that?

LoyVanBeek:
We don't have any policies or coding style guides (besides Python PEP-8) etc, but that's why we have reviews.

Issue #339, opened 2017-07-08
GPSR: ask for confirmation before leaving the arena


rokusottervanger:
To make sure the robot does not leave the arena prematurely, first ask the operator to confirm that the challenge is done.

MatthijsBurgh:
'Leave the arena' should be accepted by the action server, this way asking for confirmation is not needed.

rokusottervanger:
We should reconsider issues like this once we have next year's rulebook. Closing it for now

PR #340, opened 2017-07-08
Asking for confirmation at end of GPSR

MatthijsBurgh:
Not tested at all. Modified on GitHub

Fix for #339

### DIFF
--- a/challenge_gpsr/src/gpsr.py
+++ b/challenge_gpsr/src/gpsr.py
@@ -194,18 +194,16 @@ def main():
                     robot.speech.speak('Sorry, I will go on as long as you need me')
                     continue
                 else:
-                    pass
-            except hmi.TimeoutException:
-                # robot did not hear the confirmation, so lets go on
-                continue
-            
-            nwc = NavigateToWaypoint(robot=robot,
+                    nwc = NavigateToWaypoint(robot=robot,
                                              waypoint_designator=EntityByIdDesignator(robot=robot,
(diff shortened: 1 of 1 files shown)


challenge_gpsr/src/gpsr.py
```diff
@@ -186,12 +186,25 @@ def main():
...
+            
             nwc = NavigateToWaypoint(robot=robot,
-                                     waypoint_designator=EntityByIdDesignator(robot=robot,
-                                                                              id=knowledge.exit_waypoint),
-                                     radius = 0.3)
-            nwc.execute()
+                                             waypoint_designator=EntityByIdDesignator(robot=robot,
+                                                                                      id=knowledge.exit_waypoint),
+                                             radius = 0.3)
             robot.speech.speak("Thank you very much, and goodbye!", block=True)
```
LoyVanBeek: Should also be in the 'yes' condition

MatthijsBurgh:
@LoyVanBeek is the new version already correct?

MatthijsBurgh:
@rokusottervanger seems to work, but AMIGO is not driving back before asking if challenge is over, because that is done at the start of the loop.

MatthijsBurgh:
@rokusottervanger should we merge before Japan?

MatthijsBurgh:
I suggest to not merge this PR. Do you agree @rokusottervanger?

rokusottervanger:
I agree.

Issue #321, opened 2017-06-13
audio test fails


LoyVanBeek:
The unittests for robot_skills.test.test_audio fail now that ears.recognize has been removed.

PR #347, opened 2017-07-16
Fix audio test fails #321

LoyVanBeek:


MatthijsBurgh:
Is not crashing in sim, but needs ofcourse a test on the robot

LoyVanBeek:
Indeed, but the current version on master will for certain fail on the real robot while this yields some useful results.

PR #361, opened 2017-08-30
add zero function to arms

MatthijsBurgh:


Rayman:
You can better add it to `amigo_description` and do `amigo.arm.send_joint_goal('zero'):`

PR #369, opened 2017-09-01
Rwc2017 challenge final

Rayman:


challenge_final/launch/amigo_dependencies.launch
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
```
MatthijsBurgh: Not used in final, plus not a real feature at the moment, so could be removed in my opinion

challenge_final/scripts/challenge-final-old
```diff
@@ -1,17 +1,10 @@
 #!/usr/bin/python
```
MatthijsBurgh: Should we keep the old challenges?

challenge_final/scripts/time.py
```diff
@@ -39,7 +36,7 @@ def start_stopwatch(self):
     def get_stopwatch_time(self, stopwatch_number=0):
         if stopwatch_number >= len(self._stopwatches) or stopwatch_number < 0:
             now = datetime.now()
-            return now-now
+            return now - now
```
MatthijsBurgh: return now-now=zero? right??

challenge_final/src/challenge_final/final.py
```diff
@@ -1,31 +1,21 @@
-# ROS
-import PyKDL as kdl
 import robot_smach_states as states
 import rospy
 import smach
-
-# TU/e Robotics
-from robocup_knowledge import knowledge_loader
+from challenge_storing_groceries.manipulate_machine import PlaceSingleItem
```
MatthijsBurgh: If used in other challenges, should it be moved to smach_states?

MatthijsBurgh:
I am not sure what to do with: https://github.com/tue-robotics/tue_robocup/blob/rwc2017-challenge_final/challenge_final/src/challenge_final/pointing_detector.py#L50
It is unclear to me what code we used in the final

Rayman:
We'll fix this later

PR #371, opened 2017-09-01
Rwc2017 challenge open

Rayman:


MatthijsBurgh:
Some code is commented out. So please remove all unused code.

MaxBaeten:
As discussed with @jlunenburg and @rokusottervanger this challenge is not used anymore at least the hsr integrated part. The separate states of pointing, SSL etc could be used in future open/final challenge but this challenge is at this moment not in a testable state. (The final will be made testable since it is not integrated with HSR)

Rayman:
:+1

PR #373, opened 2017-09-01
Rwc2017 robot skills

Rayman:


robot_skills/src/robot_skills/amigo.py
```diff
@@ -1,9 +1,33 @@
...
+    def __init__(self, topic):
+        self._sub = rospy.Subscriber(topic, PoseStamped, self._callback, queue_size=1)
+        self._last_msg = None
+        self._last_received_time = rospy.Time(0)
+
+    def _callback(self, msg):
+        self._last_msg = msg
+        self._last_received_time = rospy.Time.now()
+
+    def get_last_yaw(self, max_age_seconds=2):
```
Rayman: This API is really specific to the challenge. You can better expose the callback (with a yaw for example)

robot_skills/src/robot_skills/amigo.py
```diff
@@ -1,9 +1,33 @@
...
+        q = self._last_msg.pose.orientation
+        return tf.transformations.euler_from_quaternion([q.x, q.y, q.z, q.w])[2] + math.pi
 
 
 class Amigo(robot.Robot):
+
     """docstring for Amigo"""
     def __init__(self, dontInclude=[], wait_services=False):
         super(Amigo, self).__init__(robot_name="amigo", wait_services=wait_services)
+        self.ssl = SSL("/amigo/ssl/direction_of_arrival")
```
Rayman: Hardcoded `/amigo/...`

robot_skills/src/robot_skills/amigo.py
```diff
@@ -1,9 +1,33 @@
...
 
 import robot
+import rospy
+import tf
+import math
+
+from geometry_msgs.msg import PoseStamped
+
+
+class SSL(object):
```
MatthijsBurgh: I suggest to refactor it to a robotpart. Or at least move it to somewhere else. Because it will be used on multiple robots.

robot_skills/src/robot_skills/head.py
```diff
@@ -330,7 +328,7 @@ def detect_faces(self, stamp=False):
         else:
             return self._get_faces(image).recognitions
 
-    def get_best_face_recognition(self, recognitions, desired_label):
+    def get_best_face_recognition(self, recognitions, desired_label, probability_threshold=4.0):
```
MatthijsBurgh: Needs more tuning or learning process needs to be more consistent. AMIGO was not able to recover me during following, because probability was lower than 3,5.

robot_skills/src/robot_skills/util/entity.py
```diff
@@ -146,7 +146,7 @@ def from_entity_info(e):
     super_types = e.types
 
     # TODO: this must be part of the definition of the entity in ED.
-    if e.has_shape and "amigo" not in e.id and "sergio" not in e.id and e.id is not "floor" and e.id is not "walls":
+    if e.has_shape and "amigo" not in e.id and "sergio" not in e.id and e.id != "floor" and e.id != "walls":
```
MatthijsBurgh: On rwc2017 we had wall_240 and other lengths, so better to ```"wall" not in e.id```. But best is to move it to ED.

robot_skills/src/robot_skills/amigo.py
jlunenburg: @Rayman will handle this

robot_skills/src/robot_skills/head.py
jlunenburg: @MatthijsBurgh , can you turn this into an issue in the backlog? Is too much to process before merging

robot_skills/src/robot_skills/util/entity.py
jlunenburg: @Rayman will fix this

robot_skills/src/robot_skills/head.py
Rayman: #399

robot_skills/src/robot_skills/head.py
```diff
@@ -387,167 +386,6 @@ def clear_face(self):
         rospy.loginfo('clearing all learned faces')
         self._clear_srv()
 
-    def _get_persons(self, image):
```
LoyVanBeek: Where did all this stuff go?

Rayman: https://github.com/tue-robotics/people_detector

PR #374, opened 2017-09-01
Rwc2017 challenge help me carry

Rayman:


challenge_help_me_carry/src/challenge_help_me_carry.py
```diff
@@ -141,6 +141,28 @@ def execute(self, userdata):
 
         return navigateToWaypoint.execute()
 
+class ArmWaitForMotionDone(smach.State):
```
rokusottervanger: This state is not used, so can be removed.

PR #416, opened 2017-10-03
Fix/open door

MatthijsBurgh:


MatthijsBurgh:
Tested on AMIGO

robot_smach_states/src/robot_smach_states/startup.py
```diff
@@ -117,13 +117,15 @@ def __init__(self, robot, timeout=None):
         self.door_open = Event()
 
     def avg(self, lst):
+        lst = [point for point in lst if not math.isnan(point)]
         return sum(lst)/max(len(lst), 1)
 
     def process_scan(self, scan_msg):
         try:
             middle_index = len(scan_msg.ranges)/2  # Get the middle point
             ranges_at_center = scan_msg.ranges[middle_index-2:middle_index+2]  # Get some points around the middle
```
LoyVanBeek: Then we should also check for the middle point to not be NaN. But that's for later, this change seems OK

MatthijsBurgh: You are talking about the middle_index? That one can't be a NaN, because the length of the list doesn't change.  It can be x.5 in case of an odd number of points. I don't know how that is handled.

MatthijsBurgh: The index will always be an integer, because an integer is divided by an integer. So it is already rounded.

PR #417, opened 2017-10-07
Rwc2017 challenge spr

SamAlexandrov:


challenge_spr/README.md
```diff
@@ -6,46 +6,31 @@ Launch procedure:
...
-rosrun challenge_spr challenge_spr.py amigo
+amigo-challenge_spr
 ```
 
 Testing the detection individually:
 ```
-rosrun challenge_spr detect.py amigo
+amigo-start
+amigo-free-mode
+./detect.py amigo
```
MatthijsBurgh: use rosrun

LoyVanBeek: There is a amigo-challenge-spr alias, that should be the thing to use.

MatthijsBurgh: But not for testing

PR #428, opened 2017-10-14
ed_msgs seperated

MatthijsBurgh:


reinzor:
Hard to just approve this. This change does need a lot of manual testing since we do not really have CI up and running. So please create a clean install of the system and verify that this works by testing all robocup challenges

reinzor:
Good effort btw 👍

MatthijsBurgh:
Travis is not installing ed_msgs. I don't know why. I tried a clean install and most challenges did work. When failing, it was caused by other stuff.

PR #441, opened 2017-10-19
Convert spindle to torso

MatthijsBurgh:


LoyVanBeek:
Looks good to me. Did you run a
 ```bash
ack-grep --python 'torso'
```
in the tue-robocup package? If so, I'll approve

MatthijsBurgh:
Only in stopamigo in robot_skills there is still spindle, but this whole file is deprecated

PR #443, opened 2017-10-20
clean up of arms

MatthijsBurgh:


robot_skills/src/robot_skills/arms.py
```diff
@@ -81,18 +87,26 @@ def __init__(self, robot_name, tf_listener, side):
             visualization_msgs.msg.Marker, queue_size=10)
 
     def load_param(self, param_name):
```
jlunenburg: Should we make this a member of the RobotPart class? Similar stuff is used in arms.py

robot_skills/src/robot_skills/arms.py
```diff
@@ -341,22 +373,28 @@ def handover_to_robot(self, timeout=10):
...
+        """
         Low level method that sends a array of joint references to the arm.
 
         If timeout is defined, it will wait for timeout*len(joints_reference) seconds for the
         completion of the actionlib goal. It will return True as soon as possible when the goal
         succeeded. On timeout, it will return False.
-        '''
+        :param joints_references: list of joint configurations,
+        which should be a list of the length equal to the number of joints to be moved
+        :param timeout: timeout for each joint configuration in seconds
```
jlunenburg: In general, it might be a good idea to add to the docstring what happens in case of a timeout of 0.0

robot_skills/src/robot_skills/arms.py
```diff
@@ -16,7 +16,7 @@
 from robot_part import RobotPart
 
 
-class ArmState:
+class GripperState:
```
jlunenburg: This breaks the API, are you sure you checked all relevant packages that might use this class?

MatthijsBurgh: Yep, checked all packages. Only used in two files, which I changed.

robot_skills/src/robot_skills/arms.py
MatthijsBurgh: This is arms.py ...

robot_skills/src/robot_skills/arms.py
```diff
@@ -243,38 +264,52 @@ def send_joint_goal(self, configuration, timeout=5.0):
...
+        """
         Send a GripperCommand to the gripper of this arm and wait for finishing
-        '''
+        :param state: open or close
+        :param timeout: timeout in seconds; timeout of 0.0 is not allowed
+        :return: True of False
+        """
         goal = GripperCommandGoal()
 
         if state == 'open':
```
LoyVanBeek: Comparing to a member of GripperState (used to be ArmState) for consistency is nice

robot_skills/src/robot_skills/arms.py
```diff
@@ -341,22 +373,28 @@ def handover_to_robot(self, timeout=10):
             return False
 
     def handshake(self, timeout=10):
```
LoyVanBeek: Why do we even have this method? Could be replaced with ```arm.send_joint_trajectory('handshake', timeout)```, which is also a one-liner

MatthijsBurgh: I did a cleanup, I didn't want to remove the function. but I agree that it could be removed

PR #462, opened 2017-10-26
RobotPart.load_param, docstrings and style fixes

MatthijsBurgh:


robot_skills/src/robot_skills/robot_part.py
```diff
@@ -17,8 +18,21 @@ def __init__(self, robot_name, tf_listener):
...
+        """
+        Loads a parameter from the parameter server, namespaced by robot name
+        :param param_name: parameter name
+        :param default: default value for when parameter unavailable
+        :return: loaded parameters
+        """
+        if default is None:
+            return rospy.get_param('/' + self.robot_name + '/' + param_name)
+        else:
+            return rospy.get_param('/' + self.robot_name + '/' + param_name, default)
```
Rayman: We could also run the whole challenge in a namespace. This would require a big refactor unfortunately.

MatthijsBurgh:
Not yet, I will test before merging. No problems in sim.

PR #464, opened 2017-10-28
implement new API of tf_server

MatthijsBurgh:
merge together with tue-robotics/tf_server#7


robot_skills/src/robot_skills/robot.py
```diff
@@ -131,14 +131,6 @@ def get_arm(self, side):
         backup_side = backup_obj_dict[preferred_side]
         return preferred_side, backup_side
 
-    def get_left_gripper_pose_map(self):
```
MatthijsBurgh: not used, plus should not be part of robot.py

jlunenburg:
See my comment https://github.com/tue-robotics/tf_server/issues/6

Issue #478, opened 2017-10-31
Crash of SPR challenge


rokusottervanger:
[INFO] [1509477877.130640]: 'Could you please repeat your question?'
[INFO] [1509477879.340492]: Question: Question?, grammar: 
T[{actions : <A1>}] -> C[A1]

C[{A}] -> Q[A]

V_GUIDE -> guide | escort | ..
[INFO] [1509477879.348262]: Example: 'which is the smallest drink'
[INFO] [1509477894.380648]: Preempt finished within specified preempt_timeout [1.00]
[INFO] [1509477894.381051]: Robot did not hear you (timeout)
[ERROR] [1509477894.386188]: InvalidUserCodeError: Could not execute state 'BLUFF_GAME' of type '<challenge_spr_states.bluff_game.HearTurnAndAnswerQuestions object at 0x7f28d41e3f90>': Traceback (most recent call last):
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/smach/state_machine.py", line 247, in _update_once
    self._remappings[self._current_label]))
  File "/home/amigo/ros/kinetic/repos/https_/github.com/tue-robotics/tue_robocup.git/challenge_spr/src/challenge_spr_states/bluff_game.py", line 72, in execute
    robot.speech.speak("Please ask next question!")
NameError: global name 'robot' is not defined

[ERROR] [1509477894.386576]: Robot exited with <class 'smach.exceptions.InvalidUserCodeError'>,Could not execute state 'BLUFF_GAME' of type '<challenge_spr_states.bluff_game.HearTurnAndAnswerQuestions object at 0x7f28d41e3f90>': Traceback (most recent call last):
  File "/opt/ros/kinetic/lib/python2.7/dist-packages/smach/state_machine.py", line 247, in _update_once
    self._remappings[self._current_label]))
  File "/home/amigo/ros/kinetic/repos/https_/github.com/tue-robotics/tue_robocup.git/challenge_spr/src/challenge_spr_states/bluff_game.py", line 72, in execute
    robot.speech.speak("Please ask next question!")
NameError: global name 'robot' is not defined

MatthijsBurgh:
Fixed by #485

PR #485, opened 2017-11-04
Fixed category comparison question and Crash of SPR challenge #478. 

SamAlexandrov:
It was returning: "Whoops! I cannot compare categories!"
Now it successfully compares the categories of 2 objects.


SamAlexandrov:
Fixed Crash of SPR challenge #478

PR #499, opened 2017-11-18
hmi try/except TimeoutException

MatthijsBurgh:
plus python3 compatibility issue

### DIFF
--- a/robot_skills/src/robot_skills/api.py
+++ b/robot_skills/src/robot_skills/api.py
@@ -38,7 +38,6 @@ class Api(RobotPart):
         try:
             answer = self._client.query(description, grammar, target, timeout)
         except TimeoutException as e:
-            rospy.logdebug(e)
             if callable(self._post_hook):
                 self._post_hook()
             raise e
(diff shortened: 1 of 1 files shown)


Rayman:
@MatthijsBurgh does this solve your issue with the lights?

MatthijsBurgh:
Yep it does

robot_skills/src/robot_skills/api.py
```diff
@@ -35,7 +35,13 @@ def query(self, description, grammar, target, timeout=10):
         if callable(self._pre_hook):
             self._pre_hook()
 
-        answer = self._client.query(description, grammar, target, timeout)
+        try:
+            answer = self._client.query(description, grammar, target, timeout)
+        except TimeoutException as e:
+            rospy.logdebug(e)
```
Rayman: The log is not needed because you re-reraise, the exception can be logged on a higher level

MatthijsBurgh: Done, please review again.

PR #517, opened 2017-11-21
fix(base): check if goal is running before get_state

KevinDang74:
Fixed called get_state when no goal is running

### DIFF
--- a/robot_skills/src/robot_skills/base.py
+++ b/robot_skills/src/robot_skills/base.py
@@ -13,6 +13,7 @@ import tf
 
 from cb_planner_msgs_srvs.msg import LocalPlannerAction, OrientationConstraint, PositionConstraint, LocalPlannerGoal
 from cb_planner_msgs_srvs.srv import GetPlan, CheckPlan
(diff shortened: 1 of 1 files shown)


robot_skills/src/robot_skills/base.py
```diff
@@ -47,11 +48,13 @@ def setPlan(self, plan, position_constraint, orientation_constraint):
...
     def cancelCurrentPlan(self):
-        state = self._action_client.get_state()
-        # Only cancel goal when pending or active
-        if state ==0 or state == 1:
-            self._action_client.cancel_goal()
-            self.__setState("idle")
+        if self._goal_handle:
+            state = self._action_client.get_state()
+            # Only cancel goal when pending or active
+            if state == GoalStatus.PENDING or state == GoalStatus.ACTIVE:
```
LoyVanBeek: Ah, so much nicer without magic numbers

LoyVanBeek:
Looks good to me but needs some test on Amigo

KevinDang74:
Tested on the robot and no error warning showed up, so fix is successful.

PR #542, opened 2018-01-09
WIP: doc(gpsr): update readme

rokusottervanger:
I did an update on the README about the challenge flow. It could use some more testing information though. @timassman, could you add instructions on how to install and use the command generator once you have that set up?


timassman:
added instructions to use cmdgen

Issue #283, opened 2017-05-23
Check status of robot Bodyparts at challenge start


LoyVanBeek:
For each bodypart, report when they are idle.

Do not block on this, just report and continue, as suggested by @MaxBaeten 
After this report, the operator must take action.

MatthijsBurgh:
PR is merged, so done? @LoyVanBeek

LoyVanBeek:
The merged R is about the e-stop, not all other body parts. So not yet done but n=does not have high prio either.

LoyVanBeek:
With the SSL on a separate board that's hard to see the status of, this is bound to go wrong sometime...

MatthijsBurgh:
Checking on publishing can be tricky, because it only publishes, when a new detection is received.

LoyVanBeek:
I know. I just check if there is at least one publisher on the topic, not counting messages

PR #546, opened 2018-01-13
Feature/283 check bodyparts

LoyVanBeek:
Finalizes work on #283.

Could be improved by renaming 'spindle' in the hardware diagnostics to 'torso'.

### DIFF
--- a/robot_skills/src/robot_skills/ebutton.py
+++ b/robot_skills/src/robot_skills/ebutton.py
@@ -14,9 +14,7 @@ class EButton(RobotPart):
 
         rospy.logdebug("Initializing ebutton listener")
         self._ebuttonstatus = True
-        self._topic = self.create_subscriber("/{}/emergency_switch".format(robot_name),
-                                         Bool,
-                                         self._listen)
+        self._topic = self.create_subscriber("/{}/emergency_switch".format(robot_name), Bool, self._listen)
 
     def close(self):
         pass
--- a/robot_skills/src/robot_skills/ebutton.py
+++ b/robot_skills/src/robot_skills/ebutton.py
@@ -14,7 +14,7 @@ class EButton(RobotPart):
 
         rospy.logdebug("Initializing ebutton listener")
         self._ebuttonstatus = True
-        self._topic = rospy.Subscriber("/{}/emergency_switch".format(robot_name),
+        self._topic = self.create_subscriber("/{}/emergency_switch".format(robot_name),
                                          Bool,
                                          self._listen)
 
--- a/robot_skills/src/robot_skills/robot_part.py
+++ b/robot_skills/src/robot_skills/robot_part.py
@@ -6,6 +6,7 @@ import actionlib
 
 from diagnostic_msgs.msg import DiagnosticArray
 
+
 class RobotPart(object):
     """ Base class for robot parts """
     def __init__(self, robot_name, tf_listener):
--- a/robot_skills/src/robot_skills/robot_part.py
+++ b/robot_skills/src/robot_skills/robot_part.py
@@ -46,37 +46,37 @@ class RobotPart(object):
                 return True
             # Check all connections
             new_connections = {}
-            for k, v in self.__ros_connections.iteritems():
-                # print "Checking {}".format(k)
+            for name, connection in self.__ros_connections.iteritems():
+                rospy.logdebug("Checking {}".format(name))
                 connected = False
                 # Check actionlib connection
-                if isinstance(v, actionlib.SimpleActionClient):
-                    connected = v.wait_for_server(rospy.Duration(0.01))
-                elif isinstance(v, rospy.ServiceProxy):
+                if isinstance(connection, actionlib.SimpleActionClient):
+                    connected = connection.wait_for_server(rospy.Duration(0.01))
+                elif isinstance(connection, rospy.ServiceProxy):
                     # Check service connection
                     # Need to use try-except in case of service since this throws an exception if not connected.
                     try:
-                        v.wait_for_service(timeout=0.01)
+                        connection.wait_for_service(timeout=0.01)
                         connected = True
                     except:
                         connected = False
-                elif isinstance(v, rospy.Subscriber):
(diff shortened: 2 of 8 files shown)


LoyVanBeek:
I just tested this on Amigo. 
With the e-stop pressed:

```
amigo@amigo2 $ amigo-console 

        -------------------------------------------------------------------
        | TU/e Robot Console - version 0.1                                |
        -------------------------------------------------------------------
	Succesfully loaded statemachines as "state_machine"
	Succesfully loaded designators as "ds"
	loading amigo
[ERROR] [1515846594.286495]: ... not connected timely
[WARN] [1515846599.525294]: Not all hardware operational: ['head', 'rightArm', 'leftArm', 'base']
	Succesfully loaded "amigo"
```

With the e-stop released:

```
amigo@amigo2 $ amigo-console 

        -------------------------------------------------------------------
        | TU/e Robot Console - version 0.1                                |
        -------------------------------------------------------------------
	Succesfully loaded statemachines as "state_machine"
	Succesfully loaded designators as "ds"
	loading amigo
[ERROR] [1515846614.109576]: ...  not connected timely
	Succesfully loaded "amigo"
```

LoyVanBeek:
There is one thing missing though: the torso or spindle. 
The hardware diagnostics has these items:
```bash
amigo@amigo2 $ rostopic echo /amigo/hardware_status -n 1 | grep name
    name: all
    name: base
    name: spindle
    name: left_arm
    name: right_arm
    name: head
```

So the torso has to check for 'spindle' on Amigo and for 'torso' on Sergio.

robot_skills/src/robot_skills/ebutton.py
```diff
@@ -14,7 +14,7 @@ def __init__(self, robot_name, tf_listener):
 
         rospy.logdebug("Initializing ebutton listener")
         self._ebuttonstatus = True
-        self._topic = rospy.Subscriber("/{}/emergency_switch".format(robot_name),
+        self._topic = self.create_subscriber("/{}/emergency_switch".format(robot_name),
```
jlunenburg: Indentation

robot_skills/src/robot_skills/robot_part.py
```diff
@@ -4,6 +4,7 @@
 import rospy
 import actionlib
 
+from diagnostic_msgs.msg import DiagnosticArray
```
jlunenburg: Whitespace

robot_skills/src/robot_skills/robot_part.py
```diff
@@ -99,10 +109,71 @@ def create_service_client(self, name, srv_type):
...
+        Start to check if the bodypart is operational. To do so, subscribe to the hardware status/diagnostics
+
+        :param name: check on the level of the diagnostic_msgs/DiagnosticStatus with this name
+        """
+
+        self._operational = False
+
+        self.__diagnostics_name = name
+
+        self._hardware_status_sub = self.create_subscriber("/" + self.robot_name + "/hardware_status",
```
jlunenburg: So each body part creates a subscriber to the same topic? That seems a lot of overhead. I would suggest to make this part of the Robot object, and have the callback function relay the relevant part of the message to the various body parts. Another advantage would be that you don't have to call the start_check_operational method in each (relevant) body part. Only thing is that you would have to consider when the check is required

robot_skills/src/robot_skills/robot_part.py
```diff
@@ -99,10 +109,71 @@ def create_service_client(self, name, srv_type):
...
+
+    def cb_hardware_status(self, msg):
+        """
+        hardware_status callback to determine if the bodypart is operational
+        :param msg: diagnostic_msgs.msg.DiagnosticArray
+        :return: no return
+        """
+        diags = [diag for diag in msg.status if diag.name == self.__diagnostics_name]
+
+        if len(diags) == 0:
```
jlunenburg: Is there a message definition that you could use instead of the magic 0, 1, 2? It seems a bit odd that a different convention is used than conventional (
# Possible levels of operations
byte OK=0
byte WARN=1
byte ERROR=2
byte STALE=3
)

robot_skills/src/robot_skills/torso.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, robot_name, tf_listener):
                                                               control_msgs.msg.FollowJointTrajectoryAction)
 
         # Init joint measurement subscriber
-        self.torso_sub = rospy.Subscriber('/' + self.robot_name + '/sergio/torso/measurements', JointState,
+        self.torso_sub = self.create_subscriber('/' + self.robot_name + '/sergio/torso/measurements', JointState,
```
jlunenburg: Indentation

LoyVanBeek: It's either going over the max line width or having these weird breaks.

robot_skills/src/robot_skills/robot_part.py
LoyVanBeek: I agree, though I merely moved this code. Apparently the sender is also using the values directly, not the enum value names.

LoyVanBeek:
Still working as expected:
```
Not all hardware operational: ['head', 'rightArm', 'leftArm', 'base', 'torso']
```
when E-stop is pressed, nothing if E-stop released

jlunenburg:
Additional tests:
* After booting everything incl. SSL and Jetson, reboot amigo1. Start a challenge and see if it is detected that the ROS connection is most probably not working.

LoyVanBeek:
Checking for SSL board but not yet for the Jetson board.

LoyVanBeek:
Note to self:

```python
import rosnode
rosnode.rosnode_ping("/amigo/pose_detector/openpose", max_count=1)
```

LoyVanBeek:
@MatthijsBurgh I know, I'm not an idiot AFAIK. This was a note to **self**

LoyVanBeek:
@jlunenburg There is no good way yet to check on the Jetson. I propose we add that later and merge this PR already and add the Jetson stuff later.

PR #547, opened 2018-01-13
Add NavigateToRoom

MatthijsBurgh:
This simplifies the use of navigating to a room.


robot_smach_states/src/robot_smach_states/navigation/navigate_to_symbolic.py
```diff
@@ -70,3 +70,19 @@ def generateConstraint(self):
         oc = OrientationConstraint(look_at=look_at, frame=entity_lookat.pose.frame_id)
 
         return pc, oc
+
+
+class NavigateToRoom(NavigateToSymbolic):
```
jlunenburg: You only have a docstring for the constructor, not for the class. This is not very nice.

Furthermore, I'm not too fond of the variable names. The NavigateToSymbolic class needs a dict mapping entity designators to 'area' names. Although this does not have a nice name, this makes it even more confusing.

PR #549, opened 2018-01-14
Rename EntityByIdDesignator to EdEntityByIdDesignator

MatthijsBurgh:
Done by:
```
sed -i -e 's/EntityByIdDesignator/EdEntityByIdDesignator/g' $(grep -Ril EntityByIdDesignator --exclude-dir='.*')
```
in the trunk.

### DIFF
--- a/challenge_demo/src/demo.py
+++ b/challenge_demo/src/demo.py
@@ -15,7 +15,7 @@ from action_server import Client as ActionClient
 
 from robot_smach_states.navigation import NavigateToObserve, NavigateToWaypoint, NavigateToSymbolic
 from robot_smach_states import StartChallengeRobust, WaitForTrigger
-from robot_smach_states.util.designators import EntityByIdDesignator
+from robot_smach_states.util.designators import EdEntityByIdDesignator
 
 from robot_skills.util.kdl_conversions import FrameStamped
 from robocup_knowledge import load_knowledge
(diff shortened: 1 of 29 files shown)


LoyVanBeek:
Perhaps we should take this the otherway around: remove the Ed-prefix from our designators and not tie the Entity to the world model implementation. Apparently, we're going to use WIRE again and then it makes sense to call Entities just Entities, not EdEntities.

LoyVanBeek:
This was solving https://github.com/tue-robotics/tue_robocup/pull/549

LoyVanBeek:
This was solving https://github.com/tue-robotics/tue_robocup/pull/548

MatthijsBurgh:
In the end it will still be merged to ED, so it will stay ED entities. So I would reopen this.

LoyVanBeek:
The point is still valid: Ideally I keep that we use ED as an implementation detail not exposed in class-names that should be more general.

alberth:
Ideally, you'd have a generic `Entity` interface, and an `EdEntity` implementation of it. Unfortunately, Python isn't very good at making such differences. Java would be better in this aspect.

Another solution direction could be
``` python
from robot_smach_states.util.designators import EdEntityDesignator as EntityDesignator
```
except this is hard renaming rather than the subtle difference between interface and implementation.

What strikes me as weird though is that `robot_smach_states` makes an Ed object. You'd expect that Ed performs such a function. However, I can't see how much trouble that would cause.

Maybe designators is a concept on its own, which Ed happens to store, and which smach_states happens to create?

LoyVanBeek:
> Maybe designators is a concept on its own, which Ed happens to store, and which smach_states happens to create?
indeed. Designators store objects, which can be Entities. ED happens to deliver ED messages which are converted to Entities. Wire message could also be converted to Entities, if need be. 

When Entity would be an interface, the underlying objects could be a ED-specific implementation but robot_smach_states would know only the interface and not the implementations.

PR #553, opened 2018-01-21
Fix CI for repo with multiple pkgs

MatthijsBurgh:


MatthijsBurgh:
Both builds are exceeding travis time limit

PR #556, opened 2018-01-23
Removed unnecessary semicolon

fibonatic:


LoyVanBeek:
If all is well, roslib isn't even used, so the whole import can be removed. roslib is deprecated IIRC.

fibonatic:
Just checked this in the simulator, when removing "roslib" it still works, so it seems that the import can be removed.

Issue #193, opened 2017-03-14
Cleanup kdl_conversions


jlunenburg:
- [ ] CamelCased function names
- [ ] Two functions that create a FrameStamped from x, y, z, roll, pitch, yaw, frame_id

LoyVanBeek:
As far as I can see, there is only one thing that creates

-  a Frame from x, y, z, roll, pitch, yaw (https://github.com/tue-robotics/tue_robocup/blob/master/robot_skills/src/robot_skills/util/kdl_conversions.py#L203)
-  and a FrameStamped from x, y, z, roll, pitch, yaw, frame_id (https://github.com/tue-robotics/tue_robocup/blob/master/robot_skills/src/robot_skills/util/kdl_conversions.py#L225) 

Or do you means something else

Rayman:
The CamelCased function names is still an issue

PR #557, opened 2018-01-23
193 fix kdl camel case

alberth:
Renaming all kdlCamelCase functions to kdl_camel_case functions


alberth:
Fixes the camel case problems mentioned in #193

LoyVanBeek:
I tested with this with all challenges in simulation and those all run without a problem caused by this PR.

alberth:
Much thanks for your review work

PR #562, opened 2018-02-03
Added link to wiki/Starting-the-robots

fibonatic:


MatthijsBurgh:
I don't think this should be here.

PR #565, opened 2018-02-03
cleanup()Cleaned up imports and sped up amigo-console

jlunenburg:
amigo-console contains less imports (ones that are not commonly used
have been removed).

Furthermore, the imports in all robot_skills and robot_smach_states have
been structured and unused imports have been removed.


LoyVanBeek:
Crikey, she's a big one! 

A lot of files have their hashbang ```#! usr/bin/env python``` removed but I think they are still marked as executable. To be complete, we should unmark them as such. 

I'll try to test this on Tuesday

jlunenburg:
Good point. I won' be there tomorrow so could you maybe tackle this?

Tnx!

robot_skills/src/robot_skills/head.py
```diff
@@ -1,13 +1,12 @@
 #! /usr/bin/env python
```
LoyVanBeek: Why aren't all hashbangs removed?

LoyVanBeek:
Initial tests in simulation seems to be OK

Issue #572, opened 2018-03-14
CanTransform timeout error when transforming region of interest to map


JosjaG:
When attempting to transform a kinect image to the map, the transform receives a timeout error and everything dies: 
``` 
[INFO] [1520965397.318223]: project_rois response: points: 
  - 
    header: 
      seq: 0
      stamp: 
        secs: 1520965397
        nsecs:  97251654
      frame_id: /amigo/top_kinect/openni_rgb_optical_frame
    point: 
      x: -0.409936810987
      y: -0.0194662409335
      z: 2.31227111816
Transforming roi to map
[ERROR] [1520965398.328980]: head.project_roi failed: . canTransform returned after 1.01006 timeout was 1.
```

From code:
``` 
rospy.loginfo("best_detection = {}".format(best_detection))
if best_detection:
   roi = best_detection.roi
   try:
      operator_pos_kdl = self._robot.perception.project_roi(roi=roi, frame_id="map")
   except Exception as e:
       rospy.logerr("head.project_roi failed: %s", e)
       return 'Failed'
```
The error always occurs in real life testing, and only sometimes in simulation.

LoyVanBeek:
Complete example for ```amigo-console```

```python
raw_detections = amigo.perception.detect_faces()
best_detection = amigo.perception.get_best_face_recognition(raw_detections, "operator")

rospy.loginfo("best_detection = {}".format(best_detection))
if best_detection:
   roi = best_detection.roi
   try:
      operator_pos_kdl = amigo.perception.project_roi(roi=roi, frame_id="map")
   except Exception as e:
       rospy.logerr("head.project_roi failed: %s", e)
       return 'Failed'
```

LoyVanBeek:
The call that fails is ```robot.perception.project_roi```` The simplest way to call that is with (in ```amigo-console```)
```
from sensor_msgs.msg import RegionOfInterest

amigo.perception.project_roi(RegionOfInterest(593, 619, 149, 128, False), "/map")
```

LoyVanBeek:
Call stack: 

1. https://github.com/tue-robotics/tue_robocup/blob/master/robot_skills/src/robot_skills/util/kdl_conversions.py#L71
2. https://github.com/ros/geometry/blob/indigo-devel/tf/src/tf/listener.py#L71 
3. https://github.com/ros/geometry/blob/indigo-devel/tf/src/tf/listener.py#L74 
[...]

PR #580, opened 2018-04-17
Revert "changed tf listener"

MatthijsBurgh:
This reverts commit 2a03582dbd69919fddd1d45898913581d9b098da. This shouldn't be in master. Because rips etc is broken, by this.

### DIFF
--- a/robot_skills/src/robot_skills/robot.py
+++ b/robot_skills/src/robot_skills/robot.py
@@ -19,7 +19,7 @@ import ebutton
 import lights
 
 # tf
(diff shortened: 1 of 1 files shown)


LoyVanBeek:
Relates to #572 Needs testing to show it that actually helps. @JosjaG @LarsJanssenTUe could you try with this?

LoyVanBeek:
@reinzor updated the tf library and made the issue actually go away, regardless of tf_server or tf_client.

Issue #581, opened 2018-04-17
Message is printed way too often on real robot because of no arms (and possibly spindle/torso)


SamAlexandrov:
[WARN] [1523979999.902044]: no diagnostic msg received for the right_arm
[WARN] [1523979999.902138]: no diagnostic msg received for the right_arm
[WARN] [1523979999.902448]: no diagnostic msg received for the left_arm
[WARN] [1523979999.902496]: no diagnostic msg received for the left_arm
[WARN] [1523979999.902764]: no diagnostic msg received for the spindle
[WARN] [1523979999.902765]: no diagnostic msg received for the spindle
[WARN] [1523980000.001558]: no diagnostic msg received for the head
[WARN] [1523980000.001561]: no diagnostic msg received for the head
[WARN] [1523980000.001923]: no diagnostic msg received for the right_arm
[WARN] [1523980000.001950]: no diagnostic msg received for the right_arm
[WARN] [1523980000.002183]: no diagnostic msg received for the left_arm
[WARN] [1523980000.002212]: no diagnostic msg received for the left_arm
[WARN] [1523980000.002448]: no diagnostic msg received for the spindle
[WARN] [1523980000.002486]: no diagnostic msg received for the spindle

LoyVanBeek:
This is because:

1. https://github.com/tue-robotics/tue_robocup/blob/master/robot_skills/src/robot_skills/robot.py#L76 calls 
2. https://github.com/tue-robotics/tue_robocup/blob/master/robot_skills/src/robot_skills/robot.py#L196 often. That calls 
3. https://github.com/tue-robotics/tue_robocup/blob/master/robot_skills/src/robot_skills/robot_part.py#L167 which generates that message. 

The problem is that in simulation, Sergio does have e.g. arms (which it can't yet use) but because the hardware in real life misses these components, there are no diagnostic messages for those and the crresponding body parts complain that there no message for them.

LoyVanBeek:
Most correct solution would be to remove these parts from Sergio but that requires some refactoring. Some challenges then also won't work because those should check presence of the body parts they need. 

Practically, we could just ignore some body parts in Sergio's diagnostics for now.

LoyVanBeek:
There's little point to that. The code was based on the assumption that missing diagnostic info for a part was a rare occurrence, not the default.

PR #583, opened 2018-04-17
Ignore some robot parts for diagnostics

LoyVanBeek:
Should fix https://github.com/tue-robotics/tue_robocup/issues/581

### DIFF
--- a/robot_skills/src/robot_skills/robot.py
+++ b/robot_skills/src/robot_skills/robot.py
@@ -68,6 +68,9 @@ class Robot(object):
         # Reasoning/world modeling
         self.parts['ed'] = world_model_ed.ED(self.robot_name, self.tf_listener)
 
+        # Ignore diagnostics: parts that are not present in the real robot
+        self._ignored_parts = []
+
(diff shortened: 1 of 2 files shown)


LoyVanBeek:
TODO: actually ignoring some parts in sergio :man_facepalming:

MatthijsBurgh:
I was missing that part :p

robot_skills/src/robot_skills/robot.py
```diff
@@ -68,6 +68,9 @@ def __init__(self, robot_name="", wait_services=False):
         # Reasoning/world modeling
         self.parts['ed'] = world_model_ed.ED(self.robot_name, self.tf_listener)
 
+        # Ignore diagnostics: parts that are not present in the real robot
+        self._ignored_parts = []
```
alberth: As a general rule of thumb: If you search in a container, don't use a list, since on average you have to inspect half the elements in it, and for a miss you have to check every element. Instead, use a dict or a set. (like `self._ignored_parts = set([...])`)

PR #585, opened 2018-04-22
fix(rips) style fixes

MatthijsBurgh:


challenge_rips/src/rips.py
```diff
@@ -18,73 +18,78 @@
 EXIT_2 = challenge_knowledge.exit_2
 EXIT_3 = challenge_knowledge.exit_3
 
-def setup_statemachine(robot):
 
-    sm = smach.StateMachine(outcomes=['Done','Aborted'])
+def setup_statemachine(robot):
```
LoyVanBeek: Could just as well be a class

PR #608, opened 2018-05-12
Add check for Amigo vs Sergio. If Sergio, say you don't have arms.

Jip238:


alberth:
As a suggestion (I have little insights in this part of the code), can't you ask the robot itself if it has arms, and output something based on the answer?
In that way, if we ever remove the arms of amigo, it would do the "the right thing" without any further work. (ie avoid having special cases for specific robots as much as possible).

LoyVanBeek:
@alberth that is possible. An instance of ```Robot``` has an ```.arms``` dictionary. If this is empty, the robot has no arms. An Arm-object has a ```operational``` property we might want to check for, because it looks dumb when the robot wants to use it's inoperative arms for some reason.

What makes this check difficult, however, is that we want to use Sergio in simulation with some arms so we can prepare for the actual arms to be mounted. I'm OK with the way the check is done for now. Once the arms are mounted, this script is easily updated.

challenge_presentation/src/challenge_presentation/presentation.py
```diff
@@ -11,10 +11,12 @@ class English(object):
     PURPOSE = "My purpose is to help people in domestic or care environments."
     IM_OMNIDIR = "I have an omnidirectional base instead of legs"
     EXPLAIN_BASE = "With this base, I can instantly move in any direction and turn around whenever I want"
+    NO_ARMS = "Sadly I don't have any arms."
```
jlunenburg: Yet?

challenge_presentation/src/challenge_presentation/presentation.py
```diff
@@ -29,10 +31,12 @@ class Dutch(object):
     PURPOSE = "Mijn doel is mensen te helpen in huis- en zorgomgevingen"
     IM_OMNIDIR = "In plaats van benen heb ik een omni-directioneel onderstel"
     EXPLAIN_BASE = "Met dit onderstel kan ik direct alle kanten op bewegen en omdraaien wanneer ik wil"
+    NO_ARMS = "Helaas heb ik geen armen."
```
jlunenburg: Nog niet

challenge_presentation/src/challenge_presentation/presentation.py
```diff
@@ -86,23 +90,30 @@ def execute(self, userdata=None):
...
+            function_list.append(partial(self.robot.speech.speak, self.trans.NO_ARMS, language=self.language, voice=self.voice, block=False))
+
 
         # Torso
-        function_list.append(partial(self.robot.speech.speak, self.trans.TORSO, language=self.language, voice=self.voice, block=False))
+        if self.robot.robot_name == "amigo":
+            function_list.append(partial(self.robot.speech.speak, self.trans.TORSO, language=self.language, voice=self.voice, block=False))
+        else:
+            function_list.append(partial(self.robot.speech.speak, self.trans.TORSO_SERGIO, language=self.language, voice=self.voice, block=False))
         function_list.append(partial(self.robot.torso.medium))
```
jlunenburg: Dit is ook voor SERGIO nog helemaal niet geimplementeerd, toch (d.w.z., de torso gaat helemaal niet bewegen...)? Wellicht ook even uitcommenten...

challenge_presentation/src/challenge_presentation/presentation.py
```diff
@@ -86,23 +90,30 @@ def execute(self, userdata=None):
...
-        function_list.append(partial(self.robot.leftArm.reset))
-        function_list.append(partial(self.robot.rightArm.reset))
-        function_list.append(partial(self.robot.rightArm.wait_for_motion_done))
+        if self.robot.robot_name == "amigo":
+            function_list.append(partial(self.robot.speech.speak, self.trans.TWO_ARMS, language=self.language, voice=self.voice, block=False))
+            function_list.append(partial(self.robot.speech.speak, self.trans.HUMAN_ARMS, language=self.language, voice=self.voice, block=False))
+            function_list.append(partial(self.robot.leftArm.send_joint_trajectory, "wave_front"))
+            function_list.append(partial(self.robot.speech.speak, self.trans.END_OF_ARMS, language=self.language, voice=self.voice, block=False))
+            function_list.append(partial(self.robot.speech.speak, self.trans.GRIPPERS, language=self.language, voice=self.voice, block=False))
+            function_list.append(partial(self.robot.leftArm._send_joint_trajectory, [[0, 0, 0, 1.7, 0, 0, 0]]))
```
jlunenburg: `_send_joint_trajectory` (leading underscore) is private (bewust: paden hardcoden zorgt ervoor dat dit niet per definitie transferrable is tussen de ene en de andere robot). Ik vind het dus niet acceptabel om dit te gebruiken (naar ik van Rokus begreep crasht het hier ook op omdat dit in het 'fake' object niet is geimplementeerd.

challenge_presentation/src/challenge_presentation/presentation.py
```diff
@@ -86,23 +90,30 @@ def execute(self, userdata=None):
...
+        else:
+            function_list.append(partial(self.robot.speech.speak, self.trans.NO_ARMS, language=self.language, voice=self.voice, block=False))
+
 
         # Torso
-        function_list.append(partial(self.robot.speech.speak, self.trans.TORSO, language=self.language, voice=self.voice, block=False))
+        if self.robot.robot_name == "amigo":
+            function_list.append(partial(self.robot.speech.speak, self.trans.TORSO, language=self.language, voice=self.voice, block=False))
+        else:
+            function_list.append(partial(self.robot.speech.speak, self.trans.TORSO_SERGIO, language=self.language, voice=self.voice, block=False))
```
alberth: The 'if/else' statement makes the `TORSO_SERGIO` valid for all other robots, yet the name suggests it's only for Sergio. I'd suggest to either fix the name (but that would end up near `TORSO` which apparently is also not generic, but Amigo specific, which thus seem to lead to more `TORSO`-renaming), or insert an assert that the robot is really Sergio, so it will at least crash when you try to run this at a third robot without fixing the names.

jlunenburg:
@MaxBaeten this hasn't been tested on the robot yet. If you have any time in Montreal, could you check (and merge if it works)? Tnx!

MatthijsBurgh:
This is already fixed in some way.

PR #610, opened 2018-05-15
SPR: enhancement

SamAlexandrov:
It should always be tested in the robotics_testlabs environment.


Rayman:
Let's wait for #590 to be merged

MatthijsBurgh:
So, then we should pack a scale for montreal?

PR #612, opened 2018-05-27
Feature/restaurant rotate head

### LINKED ISSUE #600: Move head while waiting for a waving person
See how to implement this: maybe wait for a while before moving the head. Definitely make it optional so that it's easy to change

MatthijsBurgh:
Please test the behaviour on a real robot.

I don't know if I should merge in RGO2018 or master. The branch in based on RGO2018.

Fixes #600


LoyVanBeek:
@LarsJanssenTUe this PR seems to be larger than it needs to be with old RGO2018 commits included as well, so can't review it in this state.
Maybe cherry-pick that commit that this is about onto the robocup branch we want to use.

PR #614, opened 2018-05-29
Rgo2018 cleanup

jlunenburg:


robot_smach_states/src/robot_smach_states/human_interaction/learn_operator.py
```diff
@@ -0,0 +1,78 @@
...
+                    if self._robot.perception.learn_person(self._operator_name):
+                        rospy.loginfo("Succesfully detected you %i times" % (num_detections + 1))
+                        num_detections += 1
+                    elif (rospy.Time.now() - learn_person_start_time).to_sec() > self._learn_person_timeout:
+                        self._robot.speech.speak("Please stand in front of me and look at me")
+                        operator = None
+                        break
+            r.sleep()
+        rospy.loginfo("We have a new operator: %s" % operator.id)
+        self._robot.speech.speak("Gotcha! I will follow you!", block=False)
```
MatthijsBurgh: This state could be used in more cases than following, therefore I would prefer to replace this line with something generic.

PR #618, opened 2018-06-12
Fix/remove cabinet inspection

KevinDang74:
removed cabinet inspection, script only works when used in the rgo2018 environment.


LoyVanBeek:
Looks and work OK. I merged this into the ```robocup``` branch so it will be merged via that way. Closing this PR into master.

PR #619, opened 2018-06-16
Help me carry sergio2

Rayman:


LarsJanssenTUe:
After robocup the old versions of the follow state and the help me carry challenge have to be kicked out, however for now this is the best solution

PR #625, opened 2018-08-01
Fix/initialize robot unspecific

JosjaG:
The initialize state called in challenges is very robot specific (it resets/restarts a few hardware and software components), which is not as it should be. This way, it does not work (and will never work) on Hero since it resets and restarts various components by name (so hardcoded) that Hero does not have. By adding a **required** reset function for each robot part, all existing parts can simply be reset when the initialize state is called. That approach works for different robots (f.e. Amigo and Hero). It is tested on Amigo and works.


robot_skills/src/robot_skills/robot_part.py
```diff
@@ -190,3 +193,10 @@ def process_hardware_status(self, diagnostic_dict):
             else:
                 self._operational = True
 
+    @abc.abstractmethod
```
LoyVanBeek: Oooh, fancy, metaclasses!
But if there is an reset-method that is inherited by all RobotPart-subclasses, why implement it everywhere? I'd say implement it only when the default ```pass``` implementation should be overridden.

rokusottervanger: So you're saying "Don't define `reset` as an abstract method, but implement it in the base class as `pass`, and override if necessary in the child classes"?

robot_smach_states/src/robot_smach_states/utility.py
```diff
@@ -20,17 +20,18 @@ def __init__(self, robot=None):
...
-        self.robot.base.local_planner._action_client.cancel_all_goals()
-        self.robot.head.reset()
-        self.robot.leftArm.reset()
-        self.robot.leftArm.send_gripper_goal('close', 0.0)
-        self.robot.rightArm.reset()
-        self.robot.rightArm.send_gripper_goal('close', 0.0)
-        self.robot.ed.reset()
-        self.robot.torso.reset()
-        self.robot.hmi.restart_dragonfly()
+        # self.robot.lights.set_color(0, 0, 1)  # be sure lights are blue
```
LoyVanBeek: Are all these outcommented lines implemented in ```reset```s of the resp. parts?

JosjaG: Yes, so every robot part that is initialized in the 'old' function now has a reset function with the same functionality.

robot_skills/src/robot_skills/robot_part.py
JosjaG: Sounds good! We implemented the suggestion to '''pass''' in the default function, and only override this where necessary, on Hero this works, but on Amigo it doesn't so we'll try to check where this goes wrong.

robot_smach_states/src/robot_smach_states/utility.py
LoyVanBeek: Aha, lights.py already had a [```reset```](https://github.com/tue-robotics/tue_robocup/blob/master/robot_skills/src/robot_skills/lights.py#L52) method, that's why I missed it here.

JosjaG:
It is tested in simulation on Amigo and on Hero on the real robot and it works in both cases.

LoyVanBeek:
Fine to merge then I think. @rokusottervanger ?

robot_skills/src/robot_skills/perception.py
```diff
@@ -34,9 +34,6 @@ def __init__(self, robot_name, tf_listener):
     def close(self):
         pass
 
-    def reset(self, timeout=0):
-        pass
-
```
MatthijsBurgh: Dit is toch overbodig

robot_skills/src/robot_skills/base.py
```diff
@@ -139,6 +139,9 @@ def computePathLength(self, path):
                 distance += math.sqrt( dx*dx + dy*dy)
         return distance
 
+    def reset(self):
```
MatthijsBurgh: Dit is toch overbodig? Local planner is ook een robotpart.

JosjaG: Yes, bij deze aangepast!

JosjaG:
Implemented the return of a boolean instead of ```pass```

robot_skills/src/robot_skills/world_model_ed.py
```diff
@@ -174,7 +174,7 @@ def get_entity_info(self, id):
 
     def reset(self, keep_all_shapes=True):
         try:
-            self._ed_reset_srv(keep_all_shapes=keep_all_shapes)
+            return self._ed_reset_srv(keep_all_shapes=keep_all_shapes)
         except rospy.ServiceException, e:
             rospy.logerr("Could not reset ED: {0}".format(e))
```
MatthijsBurgh: In case of the excption, it should return False.

JosjaG: It originally did not return anything, but I'll change it anyway

MatthijsBurgh: Old code is not always good code

robot_smach_states/src/robot_smach_states/utility.py
```diff
@@ -20,17 +20,18 @@ def __init__(self, robot=None):
...
+        #
```
MatthijsBurgh: Please remove the commented code to keep it clean

MatthijsBurgh:
@LoyVanBeek @JosjaG @PetervDooren One of you please review my changes.

PR #630, opened 2018-08-22
Fix handover operator

JosjaG:
Temporary fix for issue 99 of the action server so that now at least the normal demo works


MatthijsBurgh:
If bring should be replaced by hand-over in more places, please include in this PR.

MatthijsBurgh:
Ping!

JosjaG:
This was merely a suggestion for demo's. Replacing this in more places is part of the merge of the robocup action server.

PR #641, opened 2018-09-22
Rwc2018/challenge storing groceries2

MatthijsBurgh:


LoyVanBeek:
The door opening bit is interesting here, the other changes (that I made) I'm not so happy with. 

Maybe we should only keep the door opening and revert to what we started with actually.

MatthijsBurgh:
@LoyVanBeek I agree with you. Could you create a PR with only the door opening?

LoyVanBeek:
Done. That is #647

PR #671, opened 2018-10-16
remove hardcoded left and right arms in challenges

### LINKED ISSUE #672: remove hardcoded left and right arms in robot_skills and challenges

alberth:
fixes #672


alberth:
Second attempt, quite happy with the overall idea, but still a number of open ends.

1. Dishwasher and storing groceries challenges have been added, which still use `_send_joint_trajectory` private functions with hard-coded numbers.
2. Dishwasher and storing groceries challenges also still use robot left/right arm.
3. A number of `ArmDesignator`s exist as an alternative means to select an arm. In particular Occupied and Unoccupied arms can not be selected with my function. Not sure how to proceed, having more than one way to select an arm seems like a bad idea to me.
4. Generalizing robot class, so the specific robot can specify more about itself could be good.
 
@LoyVanBeek  @JosjaG  some feedback would be appreciated.

robot_skills/src/robot_skills/robot.py
```diff
@@ -126,34 +126,75 @@ def tf_transform_pose(self, ps, frame):
...
+        :type  goals: container wth strings or None.
+
+        :param trajectories: Collection of joint trajectories that the arm must be able to perform.
+        :type  trajectories: container wth strings or None.
+
+        :param preferred_name: Name of the preferred arm if set. By default, this is a soft
+                requirement (as in, you may get a different arm instead if the preferred arm
+                is not suitable). By appending an exclamation mark to the name, it becomes a
+                hard requirement (you will either get the preferred arm or no arm at all).
+        :type  preferred_name: Name (string) of the preferred arm, or None if any arm will do.
```
LoyVanBeek: I'd like to see this difference make explicit, not based on string-matching. So, for example only give 'left' or 'right' in preferred arm, but if you really really need a specific arm, then use eg. the ```required``` parameter.

robot_skills/src/robot_skills/arms.py
```diff
@@ -175,6 +219,34 @@ def __init__(self, robot_name, tf_listener, side):
...
 
+    def has_gripper_type(self, gripper_type):
+        """
+        Query the arm for having the proper gripper type.
+
+        :param gripper_type: Should be constant 'yes' (we have one). At some point in the
+                             future it may be useful to have a proper type.
+        :return: Whether the arm has the requested type of gripper.
+        """
+        return gripper_type == 'yes'
```
LoyVanBeek: Maybe add an Enum with all available gripper types. Eg. `pinch` or `parallel` or `suction`

robot_skills/src/robot_skills/robot.py
```diff
@@ -126,34 +126,75 @@ def tf_transform_pose(self, ps, frame):
...
+                discarded_reasons.append("{} arm is not operational".format(armname))
+                continue
+            if gripper_type is not None:
+                if not arm.has_gripper_type(gripper_type):
+                    discarded_reasons.append("{} arm does not have gripper type '{}'".format(armname, gripper_type))
+                    continue
+            if goals is not None:
+                fail = get_first_fail(goals, arm.has_joint_goal)
+                if fail is not None:
+                    discarded_reasons.append("{} arm does not have join goal '{}'".format(armname, fail))
```
LoyVanBeek: Typo: `join`

robot_skills/src/robot_skills/robot.py
```diff
@@ -126,34 +126,75 @@ def tf_transform_pose(self, ps, frame):
...
+                    continue
+            if trajectories is not None:
+                fail = get_first_fail(trajectories, arm.has_joint_trajectory)
+                if fail is not None:
+                    discarded_reasons.append("{} arm does not have join trajectory '{}'".format(armname, fail))
```
LoyVanBeek: Typo: `join`

robot_skills/src/robot_skills/robot.py
```diff
@@ -126,34 +126,75 @@ def tf_transform_pose(self, ps, frame):
...
+                    continue
+            return arms.PublicArm(arm, gripper_type, hoals, trajectories)
```
LoyVanBeek: Typo: `hoals`

robot_skills/src/robot_skills/robot.py
```diff
@@ -126,34 +126,75 @@ def tf_transform_pose(self, ps, frame):
...
+
+        rospy.logerr("Failed to find an arm")
+        for reason in discarded_reasons:
+            rospy.logdebug("  - " + reason)
```
LoyVanBeek: Maybe `logwarn`, all in one line

robot_skills/src/robot_skills/robot.py
```diff
@@ -192,6 +233,19 @@ def __exit__(self, exception_type, exception_val, trace):
             rospy.logerr("Robot exited with {0},{1},{2}".format(exception_type, exception_val, trace))
         self.close()
 
+def get_first_fail(names, func):
```
LoyVanBeek: Can be private or protected with _

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -288,7 +288,7 @@ def __init__(self, robot):
     def execute(self, userdata):
         """ Execute hook """
         if hasattr(userdata, 'arm'):
-            arm = self._robot.get_arm(userdata.arm)[0]  # Using userdata makes sure we don't need more arm designator magic
+            arm = self._robot.get_robot_arm(gripper_type="yes", preferred=userdata.arm)
```
LoyVanBeek: Here we do require a very specific arm. Otherwise, one state part of Grab uses a different arm than the other.

challenge_presentation/src/challenge_presentation/presentation.py
```diff
@@ -52,6 +52,13 @@ def __init__(self, robot, language='en'):
         smach.State.__init__(self, outcomes=["done", "preempted"])
 
         self.robot = robot
+
+        left_trajectories = ["wave_front", "show_gripper", "point_to_laser"]
+        self.left_arm = self.robot.get_robot_arm(gripper_type='yes', trajectories=left_trajectories, preferred='left!')
```
LoyVanBeek: In this case, the gripper type doesn't matter.

challenge_presentation/src/challenge_presentation/presentation.py
```diff
@@ -52,6 +52,13 @@ def __init__(self, robot, language='en'):
...
+
+        right_trajectories = ["show_gripper", "point_to_kinect"]
+        self.right_arm = self.robot.get_robot_arm(gripper_type='yes', trajectories=right_trajectories, preferred='right!')
```
LoyVanBeek: Gripper type doesn't matter. I'd say make gripper_type default to None and let that mean 'any type of gripper will do'

robot_skills/src/robot_skills/arms.py
```diff
@@ -22,6 +22,50 @@
 # Temporary: this should be approximately 0.02 once the sensor is correctly setup
 GRASP_SENSOR_LIMITS = tuple(rospy.get_param("skills/arm/grasp_sensor/limits", [0.0025, 0.18]))
 
+class PublicArm(object):
+    """
+    Public arm interface, also checks a challenge doesn't try to use more than it asked for.
+
+    :ivar _arm: Private link to the real arm.
```
LoyVanBeek: Not sure what the convention for docs on these kind of class member variables is

robot_skills/src/robot_skills/arms.py
```diff
@@ -22,6 +22,50 @@
...
+    :ivar gripper_type: Gripper type to be used.
+    :ivar joint_goals: Joint goals to be used.
+    :ivar joint_trajectories: Joint trajectories to be used.
+    """
+    def __init__(self, arm, gripper_type, joint_goals, joint_trajectories):
+        self._arm = arm
+        self.gripper_type = gripper_type
+        self.joint_goals = joint_goals
+        self.joint_trajectories = joint_trajectories
+
```
LoyVanBeek: Add getter+setter `@property` for what the arm is occupied by. These work with the ID of an Ed Entity

robot_skills/src/robot_skills/robot.py
```diff
@@ -126,34 +126,75 @@ def tf_transform_pose(self, ps, frame):
...
-            try:
-                preferred_side = self.arms[side]
-            except KeyError:
-                print "Unknown arm side:" + str(side) + ". Defaulting to 'right'"
-                preferred_side = self.arms[self.arms.keys()[0]]
-        elif isinstance(side, arms.Arm):
-            preferred_side = side
-        else:
-            print "Unknown arm side:" + str(side) + ". Defaulting to '{0}'".format(preferred_side.side)
+    def get_robot_arm(self, gripper_type=None, goals=None, trajectories=None, preferred=None):
```
LoyVanBeek: Add parameter to get an arm that is occupied by a specific Ed Entity. 
This trumps other parameters, I think. It makes no sense to specify both `required` and `occupied_by`. 

Also, when specifying `occupied_by=None`, does that mean we don't care or that we want an unoccupied arm? 
I think that `occupied_by=False` 

Or `set()` for unoccupied, and `set("<some id>")` for a specific item and `None` for 'don't care', or this but with lists: `[], ["some id"], None`

LoyVanBeek:
Also, make leftArm and rightArm a `@property` that warns the user that that member is deprecated.

robot_skills/src/robot_skills/arms.py
alberth: It's all allowed and considered to be "variable": http://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#info-field-lists

LoyVanBeek:
@alberth can you take a look at the review comments?

alberth:
I did already and wrote some code for an improved `get_arm` function. Basically all properties can be queried as needed or desired (eg for arm name `needed_name='left'` versus `required_name='left'`).
Both forms are not always mutually exclusive, eg you can express 'I need a gripper, preferably a suction gripper', or 'I need X, Y, and Z trajectories, and P trajectory would be useful too' in this way.

What you cannot express is one of several combinations of needed/desired (eg I need a suction gripper and an X trajectory, or I need a parallel gripper and an Y trajectory). While the extension to express such requests isn't big, I think it's too far for now as the challenge also has to cope with any arm it may get from such a request.

I think I stopped at the arm resolve code, I will continue next Tuesday.

challenge_presentation/src/challenge_presentation/presentation.py
alberth: set to `GRIPPER_TYPE_GRASPING`

robot_skills/src/robot_skills/arms.py
alberth: Added various `GRIPPER_TYPE_...` constants in `arms.py`

robot_skills/src/robot_skills/robot.py
alberth: All properties can be queried using the `required_xyz` parameter in which case the request must be met. Other 'nice to have' properties can be requested using the `desired_xyz` parameter.

If you get an arm, you can query if a certain property value has been granted.

Currently you get the first arm with all the required properties, which may not be the best choice. In the future, it may be better to try the other arms too, and try to get the largest possible subset of properties (for some yet unknown notion of 'largest').

alberth:
@LoyVanBeek  @jlunenburg  How does this look?

Code hasn't been built or executed yet, as I don't have a simulator up & running.

robot_skills/src/robot_skills/arms.py
```diff
@@ -22,42 +22,144 @@
...
 GRASP_SENSOR_LIMITS = tuple(rospy.get_param("skills/arm/grasp_sensor/limits", [0.0025, 0.18]))
 
+
+# Constants for arm requirements. Note that "don't care at all" is not here, as
+# it can be expressed by not imposing a requirement (set it to None).
+
+# Specific types of gripper.
+GRIPPER_TYPE_PINCH = "gripper-type-pinch"
+GRIPPER_TYPE_PARALLEL = "gripper-type-parallel"
+GRIPPER_TYPE_SUCTION = "gripper-type-suction"
```
LoyVanBeek: Sorry to nitpick, but I'd like to group these in a Python class per enum (Python 2.7 doesn't feature true Python Enum yet, that was introduced in Python 3.4). 

Eg. change to:
```python
class GripperPrinciple(object):
    Pinch = "pinch"
    Parallel = "parallel"
    Suction = "suction"
```

LoyVanBeek: Same goes for those below. Especially since 2 groups are prefixed with the same name.

robot_skills/src/robot_skills/arms.py
```diff
@@ -22,42 +22,144 @@
...
+                 has_occupied_by, avail_joint_goals, avail_joint_trajectories):
         self._arm = arm
-        self.gripper_type = gripper_type
-        self.joint_goals = joint_goals
-        self.joint_trajectories = joint_trajectories
+        self.default_gripper_type = default_gripper_type
+        self._avail_gripper_types = avail_gripper_types
+        self._has_occupied_by = has_occupied_by
+        self._avail_joint_goals = avail_joint_goals
+        self._avail_joint_trajectories = avail_joint_trajectories
```
LoyVanBeek: I prefer clarify over saving those 4 'able' characters :-). Please use the full name

robot_skills/src/robot_skills/arms.py
```diff
@@ -66,6 +168,14 @@ def close(self):
     def reset(self, timeout=0.0):
         return self._arm.reset(timeout)
 
+    def _test_die(self, cond, message):
+        """
+        Test the condition, if it fails, die with an assertion error explaining what is wrong.
+        """
+        if not cond:
+            msg = "get_robot_arm for '{}' arm did not request '{}' access."
+            raise AssertionError(msg.format(self._arm.side, message))
```
LoyVanBeek: FYI: you can also use this syntax: 
```python
assert False, "this is false"
```

which results in 
```
AssertionError: this is false
```

robot_skills/src/robot_skills/robot.py
```diff
@@ -101,6 +119,18 @@ def __init__(self, robot_name="", wait_services=False):
             not_operational_parts = [name for name, part in self.parts.iteritems() if not part.operational]
             rospy.logwarn("Not all hardware operational: {parts}".format(parts=not_operational_parts))
 
+    # Deprecated warning for 'robot.leftArm' to avoid breaking the code.
+    @proprty
```
jlunenburg: typo?

robot_skills/src/robot_skills/robot.py
```diff
@@ -129,34 +161,136 @@ def tf_transform_pose(self, ps, frame):
...
+            # Desired objects not checked currently.
+
+            # Unlike the other properties, it is likely not exactly known what objects will be used.
+            # For this reason, it seems better not to limit objects to the set specified in required
+            # _objects and desired_objects.
+            uses_objects = (required_objects is not None or desired_objects is not None)
+
+
+            # Success!
+            return arms.PublicArm(matching_arm, matching_gripper, uses_objects,
```
jlunenburg: De `matching_gripper` wordt niet gekozen uit de mathing grippers?!

robot_skills/src/robot_skills/robot.py
```diff
@@ -129,34 +161,136 @@ def tf_transform_pose(self, ps, frame):
...
+                together with other objects. None means there are no required objects.
+        :param desired_objects: Collection of objects that the arm may have. None
+                means there are no desired objects.
+
+        :return: An Arm of the robot with the requested properties, or None.
+                Note that the arm will never do more than you requested.
+        """
+        discarded_reasons = [] # Reasons why arms are discarded.
+
+        for arm_name, arm in self.iteritems():
```
jlunenburg: `self.parts.iteritems()`?

robot_skills/src/robot_skills/robot.py
```diff
@@ -195,6 +329,36 @@ def __exit__(self, exception_type, exception_val, trace):
             rospy.logerr("Robot exited with {0},{1},{2}".format(exception_type, exception_val, trace))
         self.close()
 
+def _collect_needs_desires(needs, desires, test_func):
```
jlunenburg: Check whitespaces etc. with decent linter

robot_skills/src/robot_skills/robot.py
LoyVanBeek: Typo: ```property``` instead of `proprty` (below as well)

robot_skills/src/robot_skills/robot.py
```diff
@@ -129,34 +161,136 @@ def tf_transform_pose(self, ps, frame):
...
-                preferred_side = self.arms[self.arms.keys()[0]]
-        elif isinstance(side, arms.Arm):
-            preferred_side = side
-        else:
-            print "Unknown arm side:" + str(side) + ". Defaulting to '{0}'".format(preferred_side.side)
-
-        backup_side = backup_obj_dict[preferred_side]
-        return preferred_side, backup_side
+
+    def get_robot_arm(self, required_gripper_types=None, desired_gripper_types=None,
```
LoyVanBeek: Now we have to do `robot.get_robot_arm`, which is redundant I think. Can we change this to `get_arm`?

robot_skills/src/robot_skills/robot.py
alberth: we have 2 deprecated decorators, it seems, at `tue_robocup/robot_smach_states/src/robot_smach_states/util/designators` and at `tue_robocup/robot_skills/src/robot_skills/util/decorators.py`, we may want to unify this at some point.

robot_skills/src/robot_skills/robot.py
alberth: Actually, `self.arms`, but thanks for noticing.

robot_skills/src/robot_skills/robot.py
alberth: Code did miss a default gripper (fixed now), but no, in general an arm can have several grippers.

robot_skills/src/robot_skills/robot.py
alberth: They are functions, rather than methods, but not so sure. While the code itself is generic, how often do you look for matching a need?

However, it can be moved of course. I haven't run into any generic common set of python code yet (eg also for the decorators).

robot_skills/src/robot_skills/robot.py
```diff
@@ -129,34 +161,136 @@ def tf_transform_pose(self, ps, frame):
...
+                            required_goals=None, desired_goals=None,
+                            required_trajectories=None, desired_trajectories=None,
+                            required_arm_name=None, desired_arm_name=None,
+                            required_objects=None, desired_objects=None):
```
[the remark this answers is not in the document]
alberth: The entire 'desired' part is still somewhat very optional. Code could be improved to give better garantuees about desired properties

robot_skills/src/robot_skills/robot.py
```diff
@@ -129,34 +161,136 @@ def tf_transform_pose(self, ps, frame):
...
+
+        :return: An Arm of the robot with the requested properties, or None.
+                Note that the arm will never do more than you requested.
+        """
+        discarded_reasons = [] # Reasons why arms are discarded.
+
+        for arm_name, arm in self.iteritems():
+            matching_arm = arm
+            matching_grippers = None
+            matching_goals = None
```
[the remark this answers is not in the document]
alberth: dropped

robot_skills/src/robot_skills/robot.py
alberth: done

robot_skills/src/robot_skills/robot.py
alberth: fixed, along with usn\ing an existing deprecated wrapper

alberth:
@LoyVanBeek  @jlunenburg   updated the code with the review comments

alberth:
rebased

robot_skills/src/robot_skills/robot.py
LoyVanBeek: I think its fine to leave these here, as they are only used here and _marked as private/protected.

robot_skills/src/robot_skills/robot.py
```diff
@@ -203,6 +203,20 @@ def get_arm(self, required_gripper_types=None, desired_gripper_types=None,
...
+        # Because then you might accidentally pass a GripperType instead of a [GripperType], which is a List
+        def seq_or_none(obj): 
+            return not isinstance(obj, str) and (isinstance(obj, Sequence) or obj is None)
+        assert seq_or_none(required_gripper_types)
+        assert seq_or_none(desired_gripper_types)
+        assert seq_or_none(required_goals)
+        assert seq_or_none(desired_goals)
+        assert seq_or_none(required_trajectories)
+        assert seq_or_none(desired_trajectories)
+        assert seq_or_none(required_arm_name)
```
alberth: Arm name is a string or None
```
assert required_arm_name is None or isinstance(required_arm_name, str)
```

LoyVanBeek: I noticed :-)

LoyVanBeek:
At this moment, my only doubt is whether the API is not too much. But it's the only way to have some form of checking before executing that the robot you run a challenge on is able to execute the things you want. 
There will be complaints about users of the API that it is annoying to use, because you essentially will have to declare your variables again like in old C code. But when we want to run  on all our robots (Hero, Amigo, Sergio), we will need this.

MatthijsBurgh:
@LoyVanBeek So approving it?

MatthijsBurgh:
Approving as confirmed by @LoyVanBeek

PR #691, opened 2018-11-27
Add/siza demo clean

LoyVanBeek:
This branch is a rebased version of the add/siza-demo branch: 

- there were some reverts (of a merge even). Those commits are removed completely
- Rebased onto master (instead of branched from robocup branch)

@PetervDooren and @JosjaG can you two take a look at the changes to some of the stuf that is shared with AMIGO, eg.g changes to the smach states and decide what to do with that? Both robots should run with the same state-code.


robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -318,14 +331,19 @@ def __init__(self, robot, item, arm):
...
+                                                'arrived': 'PREPARE_GRASP'})
+
+            smach.StateMachine.add('PREPARE_GRASP', PrepareEdGrasp(robot, arm, item),
+                                   transitions={'succeeded': 'GRAB',
+                                                'failed': 'RESET_FAILURE'})
+
+#            smach.StateMachine.add('NAVIGATE_TO_GRAB', NavigateToGrasp(robot, item, arm),
+#                                   transitions={'unreachable': 'RESET_FAILURE',
+#                                                'goal_not_defined': 'RESET_FAILURE',
+#                                                'arrived': 'GRAB'})
```
JosjaG: We can keep these changes, it is necessary for Hero to navigate before going to the pre-grasp pose since the updating of the to grasp entity occurs after the navigating to grasp. So if Hero goes to the pre-grasp pose before updating the entity, he can't 'see' the entity. For Amigo this is not really a problem, so keeping these changes would be best.

robot_smach_states/src/robot_smach_states/human_interaction/find_person_in_room.py
```diff
@@ -51,7 +50,7 @@ def execute(self, userdata=None):
         look_distance = 2.0  # magic number 4
         look_angles = [f * math.pi / d if d != 0 else 0.0 for f in [-1, 1] for d in [0, 6, 4, 2.3]]  # Magic numbers
         head_goals = [kdl_conversions.VectorStamped(x=look_distance * math.cos(angle),
-                                                    y=look_distance * math.sin(angle), z=1.7,
+                                                    y=look_distance * math.sin(angle), z=1.3,
```
JosjaG: This should be set o the higher goal. 1.3 meters is too low for face detection in most cases.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -33,33 +33,44 @@ def __init__(self, robot, arm, grab_entity):
...
         if not arm:
             rospy.logerr("Could not resolve arm")
             return "failed"
         userdata.arm = arm.side
 
-        entity = self.grab_entity_designator.resolve()
-        if not entity:
-            rospy.logerr("Could not resolve grab_entity")
-            return "failed"
+        # entity = self.grab_entity_designator.resolve()
```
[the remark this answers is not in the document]
JosjaG: I have no clue why there is an extra entity update, and why the order is different here... I don't think it can do any harm to leave it like this.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -33,33 +33,44 @@ def __init__(self, robot, arm, grab_entity):
...
+        # if not entity:
+        #     rospy.logerr("Could not resolve grab_entity")
+        #     return "failed"
 
         # Open gripper (non-blocking)
-        arm.send_gripper_goal('open', timeout=0)
+        #arm.send_gripper_goal('open', timeout=0)
```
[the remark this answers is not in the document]
JosjaG: Opening the gripper is already done (multiple times) in the pre-grasp state, so I suppose it is slightly unnecessary. I don't see it doing any harm, though.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -33,33 +33,44 @@ def __init__(self, robot, arm, grab_entity):
...
 
         # Torso up (non-blocking)
-        self.robot.torso.high()
```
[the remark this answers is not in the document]
JosjaG: Can't we simply also give the torso a pre-grasp pose? Since with Hero you don't want to ```look_from_above```.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -107,15 +118,15 @@ def execute(self, userdata):
...
+        #self.robot.torso.wait_for_motion_done(cancel=True)
         arm.wait_for_motion_done(cancel=True)
 
         # This is needed because the head is not entirely still when the
         # look_at_point function finishes
         rospy.sleep(rospy.Duration(0.5))
 
         # Update the entity (position)
-        segm_res = self.robot.ed.update_kinect("%s" % grab_entity.id)
+        # segm_res = self.robot.ed.update_kinect("%s" % grab_entity.id)
```
JosjaG: With Hero you don't want to update the entity after going to the pre-grasp pose, so this part was moved to before the pre-grasp.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -192,11 +203,12 @@ def execute(self, userdata):
...
 
         # Lift
+        goal_bl = grasp_framestamped.projectToFrame(self.robot.robot_name + "/base_link", tf_listener=self.robot.tf_listener)
         # rospy.loginfo('Start lifting')
         if arm.side == "left":
-            roll = 0.3
+            roll = 0.0 #0.3
         else:
-            roll = -0.3
+            roll = 0.0 #-0.3
```
JosjaG: It has to be tested whether this also works with Amigo.

robot_smach_states/src/robot_smach_states/manipulation/place.py
```diff
@@ -163,12 +165,15 @@ def execute(self, userdata=None):
         arm.occupied_by = None
 
         # Retract
+        place_pose_bl = placement_fs.projectToFrame(self._robot.robot_name+'/base_link',
+                                                    tf_listener=self._robot.tf_listener)
```
JosjaG: It has to be tested whether this also works for Amigo.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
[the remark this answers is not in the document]
JosjaG: We could also use torso.reset(), this is almost the upper limit for Amigo and the right height for Hero.

robot_smach_states/src/robot_smach_states/navigation/navigate_to_grasp.py
```diff
@@ -38,7 +38,7 @@ def generateConstraint(self):
         if arm == self.robot.arms['left']:
             angle_offset = math.atan2(-self.robot.grasp_offset.y, self.robot.grasp_offset.x)
         elif arm == self.robot.arms['right']:
-            angle_offset = math.atan2(self.robot.grasp_offset.y, self.robot.grasp_offset.x)
+            angle_offset = math.atan2(-self.robot.grasp_offset.y, self.robot.grasp_offset.x)
```
JosjaG: This is a big difference between Amigo and Hero. I suppose this should be ```arm.grasp_offset.y``` where the left arm is the same as the negative offset of the right arm, which should be defined in Amigo and Hero description.

robot_smach_states/src/robot_smach_states/perception.py
```diff
@@ -91,25 +91,29 @@ def __init__(self, robot, entity, keep_following=False, waittime=0.0):
...
+        center_point = VectorStamped(frame_id="/"+self._entity.id)
 
-        center_point.vector.z(entity.shape.z_max)
+        center_point.vector.z(self._entity.shape.z_max)
 
         rospy.loginfo('Look at %s' % (repr(center_point)))
-        robot.head.look_at_point(center_point)
-        rospy.sleep(rospy.Duration(waittime))
+        self._robot.head.look_at_point(center_point)
+        rospy.sleep(rospy.Duration(self._waittime))
```
JosjaG: These changes are an improvement and don't change the functionality, so should be kept.

robot_smach_states/src/robot_smach_states/navigation/navigate_to_grasp.py
LoyVanBeek: We can solve this indeed by defining arm-specific grasp_offsets, so we also don't have to check for left/right. 
Merging like it is now will break for AMIGO, not changing it will break for HERO. 
I think the easiest solution for now, given that generalizing the arm indices is a separate PR, is to add a check for AMIGO and HERO and removing that check when the arms are generalized for left and right. 
It's not pretty but workable and very clear that is should be cleaned up.

LoyVanBeek:
Tested on HERO, Siza demo works like a charm.

PR #701, opened 2019-01-12
Hero fixes based on testing

JosjaG:
Fixes required for Hero to be able to do the RIPS.


robot_smach_states/src/robot_smach_states/human_interaction/human_interaction.py
```diff
@@ -95,7 +95,8 @@ def execute(self, userdata=None):
 
         try:
             answer = self._robot.hmi.query('Which option?', 'T -> ' + ' | '.join(self._options), 'T', timeout=self._timeout.to_sec())
-        except TimeoutException:
+        except (Exception, TimeoutException) as e:
```
MatthijsBurgh: Are there other exceptions than the TimeoutException that can be raised here? Than it is logical to also put Exception there.

If that is the case, putting the TImeoutException there is unnecessary, because that is also an Exception.

PR #704, opened 2019-01-23
add body parts not in generic robot constructor

MatthijsBurgh:
As it should be.


MatthijsBurgh:
@hvrooy Kijk hier ook eens naar als oefening.

alberth:
This nicely clashes with #671 where I re-organized the skills base class a bit.

Like you I also had a go at cleaning it up further as separate issue, but due to these clashes I decided to postpone that until #671 was merged.

alberth:
As for review, I don't believe copy/pasting pretty much the same code (didn't very what got changed) to each of our robots is the right answer.

MatthijsBurgh:
Not all robot have the same parts, so these shouldn't be added in the generic constructor. By moving this to the robot specific constructor, the bodyparts can be more configurable. Think in orientation of a gripper, topic names etc.
This is not yet done, but should be the next step.

LoyVanBeek:
Nice work, indeed as it should be. 

But, indeed, plz wait until #671 is merged.

robot_skills/src/robot_skills/robot.py
```diff
@@ -126,7 +94,7 @@ def publish_target(self, x, y):
         self.pub_target.publish(geometry_msgs.msg.Pose2D(x, y, 0))
 
     def tf_transform_pose(self, ps, frame):
-        output_pose = geometry_msgs.msg.PointStamped
+        # output_pose = geometry_msgs.msg.PointStamped()
```
LarsJanssenTUe: Please remove this commented code!

MatthijsBurgh: Restored old situation. As I am not sure what the possible outcomes are with TF. Also not really in the focus of this PR.

robot_skills/src/robot_skills/torso.py
```diff
@@ -10,7 +10,7 @@
 
 # TU/e Robotics
 from robot_part import RobotPart
-from .util import concurrent_util
```
LoyVanBeek: I kinda like the indication that `.util` is local import. Moveover, I'd like to add the dot-prefix to `.robot_part` as well. 

That is more explicit, and explicit is better than implicit  (try `import this` in your python interpreter)

robot_skills/src/robot_skills/sergio.py
```diff
@@ -10,11 +26,31 @@ def __init__(self, wait_services=False):
 
         self._ignored_parts = ["leftArm", "rightArm", "torso", "spindle", "head"]
 
-        # This is still very ugly, because there is a lof of double code, but atleast it is only in sergio.
-        self.parts['leftArm'] = robot.arms.FakeArm(self.robot_name, self.tf_listener, side="left")
-        self.parts['rightArm'] = robot.arms.FakeArm(self.robot_name, self.tf_listener, side="right")
+        self.add_body_part('base', base.Base(self.robot_name, self.tf_listener))
+        self.add_body_part('torso', torso.Torso(self.robot_name, self.tf_listener))
+
+        self.add_body_part('leftArm', arms.FakeArm(self.robot_name, self.tf_listener, side="left"))
```
LoyVanBeek: Can you make an issue to remove this eventually?

robot_skills/src/robot_skills/torso.py
MatthijsBurgh: My Pycharm started complaining about it. So therefore I removed it. Without dot works both in terminal and PyCharm. With dot only in terminal.

robot_skills/src/robot_skills/sergio.py
MatthijsBurgh: Done #706

robot_skills/src/robot_skills/torso.py
MatthijsBurgh: Switch to absolute imports?

LoyVanBeek: Thats what we do for all of the other import I think, so then please do absolute imports. Much clearer.

robot_skills/src/robot_skills/robot.py
```diff
@@ -290,10 +288,11 @@ def __exit__(self, exception_type, exception_val, trace):
             rospy.logerr("Robot exited with {0},{1},{2}".format(exception_type, exception_val, trace))
         self.close()
 
+
 def _collect_needs_desires(needs, desires, test_func):
     """
     :param needs: Collection of needed values. None means nothing is needed.
-    :param desireds: Collection of desired values, None means nothing is desired.
+    :param desires: Collection of desired values, None means nothing is desired.
```
alberth: Actually, `desireds` was intended, as in the collection of desired things (just like `needs` is the collection of needed things).

PR #717, opened 2019-02-12
Stop having arm part variables as normal robot part

alberth:
One should use the `robot.arms` dictionary, so the code is generic and works for every robot.

Also, killed the `FakeArm`s of Sergio, running a challenge that needs arms on a robot that has none makes no sense. If you must, make the challenge aware of lacking arms.

Very tempted to also kill the `FakeArm` class, but maybe it's useful somewhen in the future (in case of emergency hacks, probably).


LoyVanBeek:
I though we did not want to give public access to arms, right? Now with the dictionary you still can. Makes for a simpler API but not robust. 

It allows to use assumptions about arm names that do not always hold (we have robots with 0, 1, 2 arms)

alberth:
The `Robot` object needs to be able to access the arm objects for hardware access. As such the `robot.arms` dictionary is fine, and it works for any number of arms. The 'name' key is somewhat questionable, but ah well.
Currently, that same `Robot` object is handed out to challenges, and probably to everything in-between. As a result, challenges have full hardware access, and thus to the `robot.arms` dictionary.

If that is undesired, you'll need to split the `Robot` class in a public interface object, and an internal 'full hardware access' object. Less drastic is to rename `Robot.arms` to `Robot._arms`, although that doesn't actually stop people from accessing, nor does it prevent code using that from getting merged into master.
To some extent, this may not be bad. Everything is experimental, so if you really really want or need, you should be able to get the 'real' robot object. It should just be very visible that it happens.

LoyVanBeek:
Changing to `Robot._arms` makes it protected and thus the IDE will complain about using it. That's fine IMO.

alberth:
Current branch introduces the `_arms` dict.

Looked for "robot.arms" in tue_robocup, found matches with some context in `robot_arms.txt`.

[robot_arms.txt](https://github.com/tue-robotics/tue_robocup/files/2907208/robot_arms.txt)
Don't know how to deal with this, or whether it is even supposed to be where it is.

LoyVanBeek:
That's going to be a lot of work... 

Most interesting: some calculations are done based on the side of an arm. Everyone knew already that was a bad idea that would not hold up and here we are. 
@jlunenburg @MatthijsBurgh any ideas in this regard?

MatthijsBurgh:
I think an arm should be configurable. So on construction some parameters can be set. An other way is to have it read the urdf model. But this requires a good design.

jlunenburg:
I'm not sure if it's that much work if I look at the list: much is deprecated or needs work anyway...

* Challenge Dishwasher: I guess this needs to be re-implemented anyway (I'm under the impression that this was quite hardcoded for AMIGO)
* Challenge manipulation: deprecated (so just remove?)
* Console: don't know
* place.py, reset.py --> no issue?
* Navigate to grasp/place: big problem: needs refactoring for HERO
* Clear.py: don't know
* arm designator: only in example?
* challenge groceries: not relevant (remove?)
* challenge test: don't know
* challenge set a table: not relevant anymore?
* challenge help me carry: not relevant anymore?

LoyVanBeek:
- [ ] `challenge_demo/src/demo.py`
- [ ] `challenge_dishwasher/src/challenge_dishwasher/simple_grab.py`
- [ ] <strike>`challenge_help_me_carry/src/challenge_help_me_carry.py`</strike> irrelevant
- [ ] `challenge_help_me_carry/src/challenge_help_me_carry2.py`
- [ ] <strike>`challenge_manipulation/src/manipulation.py`</strike> irrelevant
- [ ] <strike>`challenge_set_a_table/src/challenge_set_a_table_states/manipulate_machine.py`</strike> irrelevant
- [ ] <strike>`challenge_storing_groceries/src/challenge_storing_groceries/manipulate_machine.py`</strike> irrelevant
- [ ] `challenge_test/src/test_states.py`
- [ ] `robot_smach_states/src/robot_smach_states/clear.py`
- [ ] `robot_smach_states/src/robot_smach_states/console-nl.py`
- [ ] `robot_smach_states/src/robot_smach_states/manipulation/place.py`
- [ ] `robot_smach_states/src/robot_smach_states/navigation/navigate_to_grasp.py`
- [ ] `robot_smach_states/src/robot_smach_states/navigation/navigate_to_place.py`
- [ ] `robot_smach_states/src/robot_smach_states/reset.py`
- [ ] `robot_smach_states/src/robot_smach_states/util/designators/arm.py`

alberth:
Addressed in #751 and #752

LarsJanssenTUe:

PR #719, opened 2019-02-13
[HOTFIX] add import for deprecated decorator

MatthijsBurgh:


MatthijsBurgh:
It was not defined in that file. It was added in a PR merged recently. But the not imported, so therefore it broke.

PR #721, opened 2019-02-26
Add dependency on hmi & conv. engine

LoyVanBeek:


MatthijsBurgh:
Please also add to CMakeLists. These are not needed, as this is a python package. But CI will fail if these are not available, which is an extra check for the installation.

PR #743, opened 2019-04-02
Fix(place action): implement backup for None designator input

LarsJanssenTUe:


PetervDooren:
Checks were not passed because this pull request ran ran on travis.org instead of travis.com. according to Matthijs.

PR #754, opened 2019-04-09
Feature/path descriptor

jlunenburg:
First implementation of the path descriptor state for one of the second round challenges. I propose to review/merge it now and continue with it at a later time.


robot_smach_states/src/robot_smach_states/human_interaction/give_directions.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def execute(self, ud):
+
+        # Get the constraints for the global planner
+        nav_contraints = NavigateToSymbolic.generate_constraint(
+            robot=self._robot,
+            entity_designator_area_name_map={self._entity_designator: "near"},
+            entity_lookat_designator=self._entity_designator
+        )
+        if nav_contraints is None:
```
LoyVanBeek: Typo!

MatthijsBurgh: But it is consistent. So code shouldn't break

PR #762, opened 2019-04-24
Fix deprecated arms

MatthijsBurgh:
Fix arms keys
Make deprecated decorators more similar

### DIFF
--- a/robot_skills/src/robot_skills/util/decorators.py
+++ b/robot_skills/src/robot_skills/util/decorators.py
@@ -12,19 +12,22 @@ def deprecated(func_or_cls):
     f is deprecated
     """
     count = [0]
+
     def wrapper(*args, **kwargs):
         count[0] += 1
         if count[0] == 1:
+            item_name = func_or_cls.__name__
+            if item_name == "__init__":
+                item_name = args[0]
             try:
-                rospy.logwarn("function or class {0} is deprecated".format(func_or_cls.__name__))
-            except:
-                print func_or_cls.__name__, 'is deprecated'
+                rospy.logwarn("function or class {0} is deprecated".format(item_name))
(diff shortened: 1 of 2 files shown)


robot_skills/src/robot_skills/util/decorators.py
```diff
@@ -35,14 +38,19 @@ def deprecated(func_or_cls):
...
-        def wrapper(*args, **kwargs):
-            RED = '\033[91m'
-            END = '\033[0m'
 
-            itemname = func_or_cls.__name__
-            if itemname == "__init__":
-                itemname = args[0]
-            rospy.logerr(RED+"Function or class {0} is deprecated. You should use {1} instead".format(itemname, replacement)+END)
+        def wrapper(*args, **kwargs):
+            count[0] += 1
```
LoyVanBeek: I think you mean `count += 1`

Why is count a list even?

MatthijsBurgh: This was already this way in `deprecated` decorator. But count as an int can't be referenced inside wrapper.

LoyVanBeek: Huh?

MatthijsBurgh: I don't know why either.

alberth: Count is
```` python
count = [0]
````
(ie a list with one integer). `count[0] += 1` increments that integer, and the next line prints the warning only for `count[0] == 1` (that is, exactly one time).

I guess it's a way to get access to the integer without a `global count`, or maybe they intended to have a count for each deprecated method or so??
As Matthijs said, it's copy/paste from (iirc) SO.

PR #763, opened 2019-04-24
add right arm to hero again. Don't know if still needed, but safest solution

MatthijsBurgh:
Otherwise KeyError on line 44.
Don't know if right arm could be removed entirely or not, therefore keeping it for now.

### DIFF
--- a/robot_skills/src/robot_skills/hero.py
+++ b/robot_skills/src/robot_skills/hero.py
@@ -14,6 +14,7 @@ class Hero(robot.Robot):
         self.add_body_part('torso', torso.Torso(self.robot_name, self.tf_listener))
 
         self.add_arm_part('leftArm', arms.Arm(self.robot_name, self.tf_listener, side="left"))
+        self.add_arm_part('rightArm', arms.Arm(self.robot_name, self.tf_listener, side="right"))
 
         self.add_body_part('head', head.Head(self.robot_name, self.tf_listener))
         self.add_body_part('perception', perception.Perception(self.robot_name, self.tf_listener,


LoyVanBeek:
If we encounter issues, we should fix them the right way. @alberth added PRs to fix the issues we could encounter and that is the way we should fix issues if we encounter them. The right way.

MatthijsBurgh:
This is an issue. Right now is raises a KeyError on line 44 where it changes the joint names. Because this is still needed.

LoyVanBeek:
Isee. But then, isn't line 44 wrong? I mean: the robot only has 1 arm, our software should match that.

MatthijsBurgh:
I don't know the current state of the software 100%. So I don't know if it already compatible with one arm. So to have a safe solution for GO. Just have two arms. Unless @alberth or anybody else can tell us if we can handle one arm.

MatthijsBurgh:
@alberth you removed the specific line in 06baa4463ace2273b0acfa67ff14f4fb6d0347d0. So can you confirm that everything works without the right arm? Or did it slip in, without proper testing?

alberth:
Patch was ready for 2 weeks but never tested before merge.

I think it was a good decision to merge, or we would have left and right arms hardcoded all over the place again in all the challenges, basically just making my all my efforts useless.

MatthijsBurgh:
The final goal is definitely to get rid of this stuff. But right now we have a tournament coming up. So we need working stuff.

I checked `rightArm` is only used in gpsr (and dishwasher). So I will fix the GPSR stuff also in this PR. And then we can remove the rightarm

LoyVanBeek:
Adding this arm back on a larger PR will make the difference with other challenges only bigger.
Since all challenges are in flux right now, the least bad time to merge an impactfull PR like the arms was before tournament, so we can fix shit during tournament. If we merge it after, shit will break again potentially. 
But, yes, I agree that this is not ideal but it never will be.

PR #764, opened 2019-04-24
Store odometer data default in home folder

MatthijsBurgh:


MatthijsBurgh:
It stores the driven distance
So it will not spam the entire mega folder.

PetervDooren:
What do we use this data for? Is it used by something or could we just as well not store it?

MatthijsBurgh:
I would like to have it so we can put in the annual report, etc.

PetervDooren:
This does not sound like it requires a lot of data. There might be a better way to get those statistics other than spamming a folder with data.
I'll approve the pull request, this is not something we need to look into right now

PR #766, opened 2019-04-25
Fix arm shit including some Hacks. But without this PR, robot is useless

MatthijsBurgh:
(Also fixes GPSR)

Requires https://github.com/tue-robotics/action_server/pull/112

### DIFF
--- a/challenge_gpsr/src/gpsr.py
+++ b/challenge_gpsr/src/gpsr.py
@@ -7,7 +7,6 @@
 import sys
 import rospy
 import random
-import json
 
 import hmi
 
@@ -20,6 +19,7 @@ from robot_smach_states.util.designators import EntityByIdDesignator
 from robocup_knowledge import load_knowledge
 
 from conversation_engine import ConversationEngine
+
 
 class ConversationEngineWithHmi(ConversationEngine):
     def __init__(self, robot, grammar, command_target, knowledge):
@@ -38,7 +38,7 @@ class ConversationEngineWithHmi(ConversationEngine):
         self.finished = False
         self.start_time = rospy.get_time()
 
-        self._tc_fuckup_time = 6.0 # The TC usually needs some time to get in position and out the way of the robot
+        self._tc_fuckup_time = 6.0  # The TC usually needs some time to get in position and out the way of the robot
 
     def _say_to_user(self, message):
         rospy.loginfo("_say_to_user('{}')".format(message))
@@ -155,8 +155,8 @@ class ConversationEngineWithHmi(ConversationEngine):
         while not rospy.is_shutdown():
             try:
                 sentence, semantics = self.robot.hmi.query(description="",
-                                                      grammar=grammar,
-                                                      target=target)
+                                                           grammar=grammar,
+                                                           target=target)
                 if not self.is_text_valid_input(sentence):
                     self._say_to_user("I don't understand what you're saying, please rephrase")
                     continue
@@ -292,14 +292,12 @@ def main():
         # Move to the start location
         robot.speech.speak("Let's see if my operator has a task for me!", block=False)
 
-
     if restart:
         robot.speech.speak("Performing a restart. So sorry about that last time!", block=False)
 
     conversation_engine._start_wait_for_command(knowledge.grammar, knowledge.grammar_target)
     rospy.spin()
 
-# ------------------------------------------------------------------------------------------------------------------------
 
 if __name__ == "__main__":
     sys.exit(main())
--- a/challenge_gpsr/src/gpsr.py
+++ b/challenge_gpsr/src/gpsr.py
@@ -97,14 +97,13 @@ class ConversationEngineWithHmi(ConversationEngine):
                                                            grammar=grammar,
                                                            target=target)
 
-
                 if not self.is_text_valid_input(sentence):
                     self._say_to_user("I don't understand what you're saying, please rephrase")
                     continue
 
                 self.timeout_count = 0
                 correct = True
-                if self.test:
+                if not self.test:
                     correct = self.heard_correct(sentence)
 
                 if correct:
@@ -137,14 +136,8 @@ class ConversationEngineWithHmi(ConversationEngine):
     def _start_wait_for_command(self, grammar, target):
         rospy.loginfo("_start_wait_for_command()")
 
-        self.robot.lights.set_color(0,0,1)  #be sure lights are blue
-
+        self.robot.reset()
         self.robot.head.look_at_standing_person()
-        self.robot.leftArm.reset()
-        self.robot.leftArm.send_gripper_goal('close',0.0)
-        self.robot.rightArm.reset()
-        self.robot.rightArm.send_gripper_goal('close',0.0)
-        self.robot.torso.reset()
 
         self.robot.speech.speak("Trigger me by saying my name, and wait for the ping.", block=True)
 
@@ -163,7 +156,7 @@ class ConversationEngineWithHmi(ConversationEngine):
 
                 self.timeout_count = 0
                 correct = True
-                if self.test:
+                if not self.test:
                     correct = self.heard_correct(sentence)
 
                 if correct:
@@ -259,6 +252,8 @@ def main():
         from robot_skills.amigo import Amigo as Robot
     elif robot_name == 'sergio':
         from robot_skills.sergio import Sergio as Robot
+    elif robot_name == 'hero':
+        from robot_skills.hero import Hero as Robot
     else:
         raise ValueError('unknown robot')
 
--- a/robot_skills/src/robot_skills/arms.py
+++ b/robot_skills/src/robot_skills/arms.py
@@ -1,3 +1,5 @@
+import time
+
 # ROS
 import rospy
 import std_msgs.msg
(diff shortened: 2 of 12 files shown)


challenge_gpsr/src/gpsr.py
```diff
@@ -97,14 +97,13 @@ def _on_request_missing_information(self, description, grammar, target):
...
 
-
                 if not self.is_text_valid_input(sentence):
                     self._say_to_user("I don't understand what you're saying, please rephrase")
                     continue
 
                 self.timeout_count = 0
                 correct = True
-                if self.test:
+                if not self.test:
```
LarsJanssenTUe: Why has this changed?

robot_skills/src/robot_skills/arms.py
```diff
@@ -126,6 +119,12 @@ def send_joint_trajectory(self, configuration, timeout=5):
                        "Specify get_arm(..., required_trajectories=['{}'])".format(configuration))
         return self._arm.send_joint_trajectory(configuration, timeout)
 
+    def send_goal(self, frameStamped, timeout=30, pre_grasp=False, first_joint_pos_only=False,
```
LarsJanssenTUe: Shouldn't pre_grasp be 'true' by default?

robot_skills/src/robot_skills/hero.py
```diff
@@ -40,7 +40,6 @@ def __init__(self, wait_services=False):
        
         #rename joint names
         self.parts['leftArm'].joint_names = self.parts['leftArm'].load_param('skills/arm/joint_names')
```
LarsJanssenTUe: Shouldn't this also be generic?

robot_skills/src/robot_skills/robot.py
```diff
@@ -70,7 +71,7 @@ def configure(self):
         # Wait for connections
         s = rospy.Time.now()
         for partname, bodypart in self.parts.iteritems():
-            bodypart.wait_for_connections(1.0)
+            bodypart.wait_for_connections(0.5)
```
LarsJanssenTUe: Why is this changed?

robot_skills/src/robot_skills/robot.py
```diff
@@ -194,16 +199,32 @@ def seq_or_none(obj):
                 for match in matches:
                     matching_grippers.update(match)
 
+            # ToDO: HACK for not specifying any requirements
```
LarsJanssenTUe: Are there issues for these hacks?

robot_skills/src/robot_skills/robot.py
```diff
@@ -184,6 +185,10 @@ def seq_or_none(obj):
 
             # Grippers
             matching_grippers = set()
+            # ToDO: HACK for not specifying any requirements
```
LoyVanBeek: +100 for this. Makes API much easier to use

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -275,29 +268,30 @@ def associate(self, original_entity):
...
-        if hasattr(userdata, 'arm'):
-            arm = self._robot.get_arm(required_gripper_types=arms.GripperTypes.GRASPING,
-                                      required_arm_name=userdata.arm.side)
-        else:
-            arm = None
-        self._robot.torso.reset()  # Move up to make resetting of the arm safer
+        arm = self.arm_designator.resolve()
+        arm.reset()
+
+        # ToDo: move up... This is AMIGO specific
```
LarsJanssenTUe: Isn't this an easy fix? If so please do it now!

robot_smach_states/src/robot_smach_states/navigation/navigate_to_grasp.py
```diff
@@ -6,28 +6,29 @@
...
 
-        check_resolve_type(arm_designator, Arm) #Check that the arm_designator resolves to an Arm
+        check_resolve_type(arm_designator, PublicArm)  # Check that the arm_designator resolves to an Arm
+
         self.arm_designator = arm_designator
         if not arm_designator:
-            rospy.logerr('NavigateToGrasp: side should be determined by entity_designator. Please specify left or right, will default to left')
-            self.arm_designator = Designator(robot.leftArm)
+            rospy.logerr('NavigateToGrasp: side should be determined by entity_designator.\
+            Please specify left or right, will default to left. This is Deprecated')
```
LarsJanssenTUe: Why left?

robot_skills/src/robot_skills/arms.py
MatthijsBurgh: I don't know. Just matching the API of the hidden arm.

robot_skills/src/robot_skills/hero.py
MatthijsBurgh: Jep, but it isn't at the moment. So please make an issue for it.

robot_skills/src/robot_skills/robot.py
MatthijsBurgh: Just to speed it up.

robot_smach_states/src/robot_smach_states/navigation/navigate_to_grasp.py
MatthijsBurgh: As it says. It is deprecated. I don't know what to put in the print. So just added the deprecated sentence.

MatthijsBurgh: If you know something better. Feel free to change it.

challenge_gpsr/src/gpsr.py
MatthijsBurgh: This was wrong. If not testing, you are in a real challenge. So you're asking for confirmation.

robot_skills/src/robot_skills/robot.py
MatthijsBurgh: #767

robot_smach_states/src/robot_smach_states/manipulation/grab.py
MatthijsBurgh: Easy hack...

robot_skills/src/robot_skills/robot.py
alberth: Except you can't have a guarantee that you have sufficient capabilities if you don't specify what you need to have.

MatthijsBurgh: I know, but It would be too much to change at the moment.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -96,7 +99,6 @@ def execute(self, userdata):
         if not arm:
             rospy.logerr("Could not resolve arm")
             return "failed"
-        userdata.arm = arm.side  # Using userdata makes sure we don't need to do any more arm entity magic
```
jlunenburg: So how do you know that the arm you used for pre-grasping is also the arm used for grasping? This is not checked in any case so you´ve made it much less robust IMHO. If you don't like the user data stuff you must somehow ensure that the designator is locked so that it *can't go wrong*. That does make the designator stuff much more difficult so I don't understand why you reverted this.

robot_smach_states/src/robot_smach_states/navigation/navigate_to_grasp.py
```diff
@@ -6,28 +6,29 @@
 import rospy
 
 # TU/e Robotics
-from cb_planner_msgs_srvs.srv import *
 from cb_planner_msgs_srvs.msg import *
-from robot_skills.arms import Arm
+from robot_skills.arms import PublicArm
```
jlunenburg: Personally, I think it would have been much nicer if the arm you import was called `Arm` and the other one `PrivateArm` (so the other way around).

alberth:
On a general note, wouldn't it be useful to not merge competition stuff into master, but use some 'competition' branch instead? That would keep regular development and last minute hacking separated, so one doesn't have to untangle the mess from master afterwards.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
MatthijsBurgh: This was also not the case before.

MatthijsBurgh:
> On a general note, wouldn't it be useful to not merge competition stuff into master, but use some 'competition' branch instead? That would keep regular development and last minute hacking separated, so one doesn't have to untangle the mess from master afterwards.

That does happen on tournaments. But right now the master was broken. Something that shouldn't have happened. So the hacking is not really related to the tournament, but to a broken master and limited time.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
jlunenburg: For sure it was for the reset (see changes between lines 290 and 300). If now for a robot with two arms the first arm gives an error in grasping, it might resolve to the second arm and reset that one. You now reverted it to the old situation that has proven to work very poorly... There was even a comment on this in the code....

MatthijsBurgh: Indeed for the reset it was. But it is better to do the locking.

alberth:
Limited time doesn't count on master, it would count on a tournament branch, so as far as I am concerned you're doing tournament hacking in master now.

PR #827, opened 2019-05-14
Challenge find my mates PR

JosjaG:


challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+        look_angles = [f * math.pi / d if d != 0 else 0.0 for f in [-1, 1] for d in [0, 6, 4, 2.3]]  # Magic numbers
+        head_goals = [kdl_conversions.VectorStamped(x=look_distance * math.cos(angle),
+                                                    y=look_distance * math.sin(angle), z=1.7,
+                                                    frame_id="/%s/base_link" % self._robot.robot_name)
+                      for angle in look_angles]
+        head_goals.extend([kdl_conversions.VectorStamped(x=2.0 * look_distance * math.cos(angle),
+                                                         y=2.0 * look_distance * math.sin(angle), z=1.7,
+                                                         frame_id="/%s/base_link" % self._robot.robot_name)
+                           for angle in look_angles])
+        n = 0
```
LoyVanBeek: What's with the `n` and `i`? Can you explain what each does?

I think for `i` cyou can just enumerate the head goals: 

```python
for i, head_goal in enumerate(head_goals):
   # do stuff
```

JosjaG: The ```i``` is copied from ```learn_operator``` (just as most of that smach state), and you want to reset ```i``` to 0 to keep looking until the timeout is reached, which is not possible using enumerate, right? 

And the ```n``` represents the number of people detected, since you want to keep looking until you've found four people. I'll describe that in a comment there.

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+            # Edit these sentences based on available attributes
+            if unique_property == 'gender':
+                sentence += "and the person was a {}.".format(getattr(person, unique_property))
+            elif unique_property == 'tall':
+                sentence += "and the person was tall. "
+            elif unique_property == 'shirt_colour':
+                sentence += "and the person was wearing a mostly {} shirt.".format(getattr(person, unique_property))
+
+            people_descriptions[getattr(person, 'name')] = sentence
+            if i > 3:
```
PetervDooren: Why not create a unique description for all people? Or are we certain that the operator will ask for descriptions of the first three people?

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+                    n += 1
+            else:
+                self._robot.speech.speak(
+                    "I don't understand, I expected a command like " + ", ".join(challenge_knowledge.names))
+                continue
+
+        return 'Done'
+
+
+class AskNames(State):
```
PetervDooren: Is this state still used?

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+                return 'Aborted'
+
+            self._robot.head.look_at_point(head_goals[i])
+            i += 1
+            if i == len(head_goals):
+                i = 0
+            self._robot.head.wait_for_motion_done()
+            raw_detections = self._robot.perception.detect_faces()
+
+            # if not raw_detections:
```
PetervDooren: These lines are no longer necessary right?

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+                            frame_stamped=kdl_conversions.FrameStamped(kdl.Frame(person_pos_kdl.vector), "map"),
+                            type="waypoint")
+
+                        n += 1
+                    if len(self._people) > 3:
+                        self._robot.head.close()
+                        return 'Done'
+
+        self._robot.head.close()
+        # rospy.sleep(2.0)
```
PetervDooren: Is this line still necessary?

challenge_find_my_mates/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
...
+
+catkin_python_setup()
+
+catkin_package()
+
+#############
+## Testing ##
+#############
+
+## Add folders to be run by python nosetests
```
MatthijsBurgh: Please remove commented code.

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+            if raw_detections:
+                for detection in raw_detections:
+                    roi = detection.roi
+                    try:
+                        person_pos_kdl = self._robot.perception.project_roi(roi=roi, frame_id="map")
+                    except Exception as e:
+                        rospy.logerr("head.project_roi failed: %s", e)
+                        continue  # move onto next detection
+
+                    detected_person = self._robot.ed.get_closest_laser_entity(radius=self._look_distance,
```
MatthijsBurgh: Add a comment that this should be changed to new people stuff in ED.

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+                                                                              center_point=person_pos_kdl)
+
+                    room_entity = self._robot.ed.get_entity(id=ROOM_ID)
+                    if not room_entity.in_volume(detected_person.pose.extractVectorStamped(), 'in'):
+                        detected_person = None
+
+                    for person in self._people:
+                        if person.pose.extractVectorStamped() - detected_person._pose.extractVectorStamped() > \
```
MatthijsBurgh: This if looks very complex.

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+                            self._min_dist and self._robot.ed.get_entities(id=STARTING_POINT).\
+                            _pose.extractVectorStamped() - detected_person.pose.extractVectorStamped() > self._min_dist:
+                            self._people[n] = detected_person
+
+                    if len(self._people) == (n+1):
```
MatthijsBurgh: If the previous IF is true, is this IF then always true? Then this IF is unnecessary.

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+    Locate three people in the provided room
+    """
+
+    def __init__(self, robot, search_timeout=60, look_distance=2.0, min_dist=0.5):
+        """ Initialization method
+        :param robot: robot api object
+        :param search_timeout: (float) maximum time the robot is allowed to search
+        :param look_distance: (float) robot only considers laser entities within this radius
+        """
+        State.__init__(self, outcomes=['Done', 'Aborted', 'Failed'])
```
MatthijsBurgh: I think it is good practice to have the return strings to be all lowercase.

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+        return 'Done'
+
+
+class IdentifyPeople(State):
+    """
+    Navigate to all three people and determine their attributes
+    """
+
+    def __init__(self, robot):
+        State.__init__(self, outcomes=['Done', 'Aborted', 'Failed'])
```
MatthijsBurgh: I think it is good practice to have the return strings to be all lowercase.

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+                command_recognized = result.sentence
+            except TimeoutException:
+                command_recognized = None
+            if command_recognized == "":
+                self._robot.speech.speak("I could not hear your name, please tell me your name.")
+            elif command_recognized in challenge_knowledge.names:
+                self._robot.speech.speak(
+                    "OK, I understand your name is {}, but I will think of a nice nickname.".format(command_recognized))
+            else:
+                self._robot.speech.speak(
```
MatthijsBurgh: This is correct code, but very weird behaviour. The robot says than you can only be named one of X.

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+                    "I don't understand, I expected a command like " + ", ".join(challenge_knowledge.names))
+            try:
+                print("Detecting this person")
```
MatthijsBurgh: Replace by roslog

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+        return 'Done'
+
+
+class ReportPeople(State):
+    """
+    Form the sentences for all three people and say that sentence
+    """
+
+    def __init__(self, robot):
+        State.__init__(self, outcomes=['Done', 'Aborted', 'Failed'])
```
MatthijsBurgh: I think it is good practice to have the return strings to be all lowercase.

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+        furniture_entities = [entity for entity in entities if entity.is_a("furniture")]
+
+        # Match the furniture entities to rooms
+        furniture_entities_room = []
+        for item in furniture_entities:
+            if self._room_entity.in_volume(kdl_conversions.VectorStamped(vector=item._pose.p), "in"):
+                furniture_entities_room.append(item)
+        closest_entity = []
+        for person in person_entities:
+            distance_closest = 10.0
```
MatthijsBurgh: Magic number

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+            temp_entity_near = furniture_entities_room[0]
```
MatthijsBurgh: Are you sure the list always contains a element? Otherwise stuff will break.

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+                    distance_closest = distance
+            closest_entity.append(temp_entity_near)
+
+        self._robot.speech.speak('I found many people but none were as pretty as you.', block=True)
+        people_descriptions = {}
+        for i, person in enumerate(person_entities):
+            sentence = "I found {} near the {}.\n".format(getattr(person, 'name'), closest_entity[i].id)
+
+            unique_property = None
+            properties = [prop for prop in dir(person) if not prop.startswith('__')]
```
MatthijsBurgh: This should be changed to contain the new person information

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+        return 'Done'
+
+
+class AskNames(State):
+    """
+    Ask for which three people to describe
+    """
+
+    def __init__(self, robot):
+        State.__init__(self, outcomes=['Done', 'Aborted', 'Failed'])
```
MatthijsBurgh: I think it is good practice to have the return strings to be all lowercase.

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,338 @@
...
+                command_recognized = None
+            if command_recognized == "":
+                self._robot.speech.speak("I am still waiting for a command and did not hear anything")
+            elif command_recognized in challenge_knowledge.common.names:
+                    self._robot.speech.speak("OK, I will describe {}".format(command_recognized), block=True)
+                    self._robot.speech.speak(people_descriptions[command_recognized], block=True)
+                    n += 1
+            else:
+                self._robot.speech.speak(
+                    "I don't understand, I expected a command like " + ", ".join(challenge_knowledge.names))
```
MatthijsBurgh: A command... Names. But again the weird robot behaviour. Plus could remove the duplicate code?

challenge_find_my_mates/src/find_my_mates.py
JosjaG: This is all people, this if is only met when there are four people (```>``` not ```=>```), and the challenge states there will be four people. But that should then also be the end of the loop so the ```break``` is unnecessary.

challenge_find_my_mates/src/find_my_mates.py
JosjaG: It will be used (as mentioned in the comment in line 152) to update the person entity to include a name.

challenge_find_my_mates/src/find_my_mates.py
JosjaG: It is, but it's also necessary, what solution would you propose?

challenge_find_my_mates/src/find_my_mates.py
JosjaG: What would your proposed solution be?

challenge_find_my_mates/src/find_my_mates.py
MatthijsBurgh: Now I also see the mixed usage of `.pose` and `._pose`.

But create some extra variables, that would already clear things up.
`get_entities` can return multiple entities, so this code can break.
There is a repeating element in it, so that could be put in a function
Add some comments about what is happening here.

JosjaG: I created extra variables with names that clarify what's going on. ```get_entities``` could indeed break, but if it would break on the starting point it will already break at the start first I suppose. I fear attempting to catch that possible error will over-complicate this. I fixed the mixed use of ```pose ``` and ```_pose```.

challenge_find_my_mates/src/find_my_mates.py
```diff
@@ -0,0 +1,308 @@
...
+            except TimeoutException:
+                command_recognized = None
+            if command_recognized == "":
+                self._robot.speech.speak("I could not hear your name, please tell me your name.")
+            elif command_recognized in challenge_knowledge.names:
+                self._robot.speech.speak(
+                    "OK, I understand your name is {}, but I will think of a nice nickname.".format(command_recognized))
+            else:
+                self._robot.speech.speak(
+                    "I don't understand, I expected a command like " + ", ".join(challenge_knowledge.names))
```
MatthijsBurgh:
```suggestion
                    "I don't understand, I expected a one of the following names: " + ", ".join(challenge_knowledge.names))
```

PR #830, opened 2019-05-18
Feature/rise for hmi state

PetervDooren:
Smach state for bringing hero into a pose for human machine interaction

### DIFF
--- a/robot_smach_states/src/robot_smach_states/rise.py
+++ b/robot_smach_states/src/robot_smach_states/rise.py
@@ -6,7 +6,7 @@ import smach
 from robot_skills.util.kdl_conversions import VectorStamped
 
 
-class HeroHMIPose(smach.State):
+class HeroHmiPose(smach.State):
     """
     State to ensure that hero's arm is out of the way so that the robot
     may move his torso up.
@@ -32,6 +32,19 @@ class HeroHMIPose(smach.State):
         arm.wait_for_motion_done()
         return "succeeded"
 
+class HmiPose(smach.State):
+    """
+    State to puts the robot in a position for hmi
+    :param robot: Robot to execute state with
+    """
+    def __init__(self, robot):
+        smach.State.__init__(self, outcomes=['succeeded', 'failed'])
+        self._robot = robot
+
+    def execute(self, _):
+        self._robot.head.look_at_standing_person()
+        return "succeeded"
+
 
 class RiseForHMI(smach.StateMachine):
     """
@@ -55,6 +68,11 @@ class RiseForHMI(smach.StateMachine):
         smach.StateMachine.__init__(self, outcomes=['done', 'failed'])
 
         with self:
-            smach.StateMachine.add('ARM_OUT_OF_WAY', HeroHMIPose(robot),
-                                   transitions={'succeeded': 'done',
-                                                'failed': 'failed'})
+            if robot.robot_name == 'hero2':
+                smach.StateMachine.add('HERO_HMI_POSE', HeroHmiPose(robot),
+                                       transitions={'succeeded': 'done',
+                                                    'failed': 'failed'})
+            else:
+                smach.StateMachine.add('HMI_POSE', HmiPose(robot),
+                                       transitions={'succeeded': 'done',
+                                                    'failed': 'failed'})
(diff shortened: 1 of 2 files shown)


robot_smach_states/src/robot_smach_states/rise.py
```diff
@@ -0,0 +1,67 @@
...
+        smach.State.__init__(self, outcomes=['succeeded', 'failed'])
+        self._robot = robot
+
+    def execute(self, _):
+        rotation = 1.57
+        rot_speed = 1
+        arm = self._robot.get_arm()
+
+        #get position to look at
+        goal = VectorStamped(1.0, 0.0, 1.6, frame_id="/" + self._robot.robot_name + "/base_link")
```
JosjaG: This is the exact same pose as ```look_at_standing_person```, why not just use ```self._robot.head.look_at_standing_person()```?

robot_smach_states/src/robot_smach_states/rise.py
```diff
@@ -0,0 +1,67 @@
...
+    State to ensure that hero's arm is out of the way so that the robot
+    may move his torso up.
+    :param robot: Robot to execute state with
+    """
+    def __init__(self, robot):
+        smach.State.__init__(self, outcomes=['succeeded', 'failed'])
+        self._robot = robot
+
+    def execute(self, _):
+        rotation = 1.57
```
JosjaG: Could it be convenient for these to be parameters that you can provide to this state? Just use these as defaults but allow for them to be provided as well?

robot_smach_states/src/robot_smach_states/rise.py
```diff
@@ -0,0 +1,67 @@
...
+
+        #get position to look at
+        goal = VectorStamped(1.0, 0.0, 1.6, frame_id="/" + self._robot.robot_name + "/base_link")
+        tf_goal = goal.projectToFrame('/map', self._robot.tf_listener)
+
+        if arm.has_joint_goal('arm_out_of_way'):
+            arm.send_joint_goal('arm_out_of_way', 0.0)
+            self._robot.base.force_drive(0, 0, rot_speed, rotation / rot_speed)
+        self._robot.head.look_at_point(tf_goal)
+        arm.wait_for_motion_done()
```
JosjaG: Should this then not also include a ```self._robot.head.wait_for_motion_done()``` before returning?

robot_smach_states/src/robot_smach_states/rise.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    State to puts the robot in a position for hmi
+    :param robot: Robot to execute state with
+    """
+    def __init__(self, robot):
+        smach.State.__init__(self, outcomes=['succeeded', 'failed'])
+        self._robot = robot
+
+    def execute(self, _):
+        self._robot.head.look_at_standing_person()
```
JosjaG: Again, add a ```self._robot.head.wait_for_motion_done()``` before returning from this state.

robot_smach_states/src/robot_smach_states/rise.py
PetervDooren: look_at_standing_person() is defined with respect to the base of the robot. Since we are turning the robot sideways we need to take the point and project it to map frame. 
We only want to move the head after turning is complete, otherwise the robot won't look at the person.

JosjaG: Ah, alright! Perhaps add a comment clarifying this :)

JosjaG: Ah, alright! Perhaps add a comment clarifying this :)

robot_smach_states/src/robot_smach_states/rise.py
PetervDooren: Agreed, done

robot_smach_states/src/robot_smach_states/rise.py
PetervDooren: Done

robot_smach_states/src/robot_smach_states/rise.py
PetervDooren: Made the parameters arguments for the HeroHmiPose state

robot_smach_states/src/robot_smach_states/rise.py
PetervDooren: Added a comment for clarification

PR #832, opened 2019-05-18
Rgo2019 challenge gpsr

### LINKED ISSUE #815: RGO2019 PR tue_robocup rgo2019_challenge_gpsr

LarsJanssenTUe:
Mostly addition of knowledge and giving clearer speech commands.

Fixes #815


challenge_gpsr/src/gpsr.py
```diff
@@ -74,7 +75,7 @@ def task_finished(self, message):
         else:
             rospy.logwarn("Not going to meeting point, challenge has param skip:=true set")
 
-        if self.tasks_done >= self.tasks_to_be_done and not self.skip:
+        if (self.tasks_done >= self.tasks_to_be_done and not self.skip) or (self.finished and not self.skip and self.tasks_done >= 1):
```
MatthijsBurgh: This becomes very complex...

challenge_gpsr/src/gpsr.py
```diff
@@ -139,19 +141,19 @@ def _start_wait_for_command(self, grammar, target):
         self.robot.reset()
         self.robot.head.look_at_standing_person()
 
-        self.robot.speech.speak("Trigger me by saying my name, and wait for the ping.", block=True)
+        self.robot.speech.speak("Please trigger me by stating my name directly into the microphone.", block=True)
```
MatthijsBurgh: I would keep the ping in there.

challenge_gpsr/src/gpsr.py
```diff
@@ -139,19 +141,19 @@ def _start_wait_for_command(self, grammar, target):
...
 
         self.wait_to_be_called()
 
-        self.robot.speech.speak("What can I do for you?", block=True)
+        self.robot.speech.speak("What can I do for you? State your command loudly directly into my microphone, but please wait for the ping!.", block=True)
```
MatthijsBurgh: Too much to repeat this every time.

challenge_gpsr/src/gpsr.py
```diff
@@ -139,19 +141,19 @@ def _start_wait_for_command(self, grammar, target):
...
 
         while not rospy.is_shutdown():
             try:
                 sentence, semantics = self.robot.hmi.query(description="",
                                                            grammar=grammar,
                                                            target=target)
                 if not self.is_text_valid_input(sentence):
-                    self._say_to_user("I don't understand what you're saying, please rephrase")
+                    self._say_to_user("I don't understand what you're saying, please rephrase. Speak loudly and clearly after the ping.")
```
MatthijsBurgh: Same

challenge_gpsr/src/gpsr.py
```diff
@@ -191,7 +193,7 @@ def wait_to_be_called(self):
                     rospy.logwarn("[GPSR] Timeout_count: {}".format(self.timeout_count))
 
     def heard_correct(self, sentence):
-        self.robot.speech.speak('I heard %s, is this correct?' % sentence)
+        self.robot.speech.speak('I heard %s, is this correct? Please speak loudly into the microphone and move away afterwards!' % sentence)
```
MatthijsBurgh: Same

challenge_gpsr/src/gpsr.py
LarsJanssenTUe: I simplified this a little bit, however this addition ensures that the robot drives out of the arena when there is only a little bit of time left. This should result in a very high chance to obtain extra points for autonomously driving out of the arena while just doing an extra assignment has a lower chance of winning more points.

challenge_gpsr/src/gpsr.py
LarsJanssenTUe: Shortened by introducing the tc_fuckup time once again

challenge_gpsr/src/gpsr.py
MatthijsBurgh: IMO it would be best to only have a finished check here. You can keep the skip here as well. But finished should be determined somewhere else. That logic can be more complex.

LarsJanssenTUe: Let's meet half way:

```
 if (self.tasks_done >= self.tasks_to_be_done or self.finished) and not self.skip:
```

MatthijsBurgh:
Please wait with merging till Travis succeeds or runs into time limit.

PR #857, opened 2019-06-08
Feature/rise for inspect

PetervDooren:
Added a state which brings the robot in the correct pose for an inspect.
Also refactored the riseforHmi state so that the robot dependent part is in the robot object.

### DIFF
--- a/robot_skills/src/robot_skills/hero.py
+++ b/robot_skills/src/robot_skills/hero.py
@@ -51,6 +51,10 @@ class Hero(robot.Robot):
         self.configure()
 
     def move_to_inspect_pose(self, inspect_target):
+        """
+        This poses the robot for an inspect.
+        :param inspection_position: kdl.Frame with the pose of the entity to be inspected.
+        """
         # calculate the arm_lift_link which must be sent
         z_over = 0.3  # height the robot should look over the surface
         z_hh = 0.9  # height of the robots head at z_arm=0
@@ -89,6 +93,9 @@ class Hero(robot.Robot):
         return True
 
     def move_to_hmi_pose(self):
+        """
+        This poses the robot for conversations.
+        """
         arm = self.get_arm()
 
         rotation = 1.57
--- a/robot_skills/src/robot_skills/hero.py
+++ b/robot_skills/src/robot_skills/hero.py
@@ -50,6 +50,44 @@ class Hero(robot.Robot):
 
         self.configure()
 
+    def move_to_inspect_pose(self, inspect_target):
+        # calculate the arm_lift_link which must be sent
+        z_over = 0.3  # height the robot should look over the surface
+        z_hh = 0.9  # height of the robots head at z_arm=0
+        torso_to_arm_ratio = 2  # motion of arm/motion of torso
+        z_head = inspect_target.z() + z_over
+
+        # check whether moving the arm is necessary
+        if z_head < 1.2:
+            rospy.loginfo('Entity is low enough. we dont need to move the arm')
+            return True
+
+        # saturate the arm lift goal
+        z_arm = (z_head - z_hh) * torso_to_arm_ratio
+        if z_arm > 0.69:
+            z_arm = 0.69
+            rospy.logwarn('Warning: looking at excessively high surface')
+        elif z_arm < 0.0:
+            z_arm = 0.0
+            rospy.logwarn('Surface is low enough, we dont need to move the arm.')
+
+        arm = self.get_arm()
+
+        # turn the robot
+        rotation = 1.57
+        rotation_speed = 1
+        rotation_duration = rotation / rotation_speed
+        if arm.has_joint_goal('arm_out_of_way'):
+            pose = arm._arm.default_configurations['arm_out_of_way']
+            pose[0] = z_arm
+            arm._arm._send_joint_trajectory([pose])
+        else:
+            rospy.logwarn('Warning: robot does not have an "arm_out_of_way" joint goal')
+            return False
+        self.base.force_drive(0, 0, rotation_speed, rotation_duration)
+        arm.wait_for_motion_done()
+        return True
+
 
 if __name__ == "__main__":
     rospy.init_node("hero")
--- a/robot_skills/src/robot_skills/hero.py
+++ b/robot_skills/src/robot_skills/hero.py
@@ -48,46 +48,45 @@ class Hero(robot.Robot):
         for arm in self.arms.itervalues():
             arm._operational = True
 
+        # verify joint goal required for posing
+        assert 'arm_out_of_way' in self.parts['leftArm'].default_configurations,\
+            "arm_out_of_way joint goal is not defined in {}_describtion skills.yaml".format(self.robot_name)
+
         self.configure()
 
     def move_to_inspect_pose(self, inspect_target):
         """
         This poses the robot for an inspect.
+
         :param inspection_position: kdl.Frame with the pose of the entity to be inspected.
+        :return result: boolean, false if something went wrong.
         """
         # calculate the arm_lift_link which must be sent
         z_over = 0.3  # height the robot should look over the surface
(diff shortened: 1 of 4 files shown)


robot_skills/src/robot_skills/hero.py
```diff
@@ -50,6 +50,63 @@ def __init__(self, wait_services=False):
...
 
+    def move_to_inspect_pose(self, inspect_target):
+        """
+        This poses the robot for an inspect.
+        :param inspection_position: kdl.Frame with the pose of the entity to be inspected.
+        """
+        # calculate the arm_lift_link which must be sent
+        z_over = 0.3  # height the robot should look over the surface
+        z_hh = 0.9  # height of the robots head at z_arm=0
+        torso_to_arm_ratio = 2  # motion of arm/motion of torso
```
jlunenburg: 2.0? If you (accidentally) use two ints when you mean two floats you will get unexpected results (
```
In [13]: 1/2
Out[13]: 0
```

robot_skills/src/robot_skills/hero.py
```diff
@@ -50,6 +50,63 @@ def __init__(self, wait_services=False):
...
+        z_head = inspect_target.z() + z_over
+
+        # check whether moving the arm is necessary
+        if z_head < 1.2:
+            rospy.loginfo('Entity is low enough. we dont need to move the arm')
+            return True
+
+        # saturate the arm lift goal
+        z_arm = (z_head - z_hh) * torso_to_arm_ratio
+        if z_arm > 0.69:
```
jlunenburg:
```python
# Limit z_arm
z_arm = min(0.69, max(0.0, z_arm))
```

robot_skills/src/robot_skills/hero.py
```diff
@@ -50,6 +50,63 @@ def __init__(self, wait_services=False):
...
+            rospy.logwarn('Warning: looking at excessively high surface')
+        elif z_arm < 0.0:
+            z_arm = 0.0
+            rospy.logwarn('Surface is low enough, we dont need to move the arm.')
+
+        arm = self.get_arm()
+
+        # turn the robot
+        rotation = 1.57
+        rotation_speed = 1
```
jlunenburg: 1.0

robot_skills/src/robot_skills/hero.py
```diff
@@ -50,6 +50,63 @@ def __init__(self, wait_services=False):
...
+        rotation_duration = rotation / rotation_speed
+        if arm.has_joint_goal('arm_out_of_way'):
```
jlunenburg: I would check this in the `__init__`:
```python
assert arm.has_joint_goal('arm_out_of_way`), "arm_out_of_way joint goal is not defined in ..."
```
This way, you know on startup that something's missing instead of halfway through your challenge.

robot_skills/src/robot_skills/hero.py
```diff
@@ -50,6 +50,63 @@ def __init__(self, wait_services=False):
...
+        arm = self.get_arm()
+
+        rotation = 1.57
+        rotation_speed = 1
+        rotation_duration = rotation / rotation_speed
+        if arm.has_joint_goal('arm_out_of_way'):
+            arm.send_joint_goal('arm_out_of_way', 0.0)
+            self.base.force_drive(0, 0, rotation_speed, rotation_duration)
+        arm.wait_for_motion_done()
+        return "succeeded"
```
jlunenburg: Why return a 'succeeded' string? I'd use a bool *if* you check for the return values of base and arm motions and otherwise simply don't return anything.

Make sure to also add this to your docstring.

robot_skills/src/robot_skills/robot.py
```diff
@@ -314,6 +314,21 @@ def handle_hardware_status(self, diagnostic_array):
             if name not in self._ignored_parts:
                 part.process_hardware_status(diagnostic_dict)
 
+    def move_to_inspect_pose(self, inspection_position):
+        """
+        This poses the robot for an inspect.
```
jlunenburg: Specify in you docstrings what these methods should return!

* Either: enforce that inheriting robots implement these methods by raising `NotImplementedErrors`
* Or: don't care whether they do; in that case: don't print this (either no log statements or only debug statements)

robot_smach_states/src/robot_smach_states/world_model.py
```diff
@@ -203,12 +204,16 @@ def __init__(self, robot, entityDes, objectIDsDes=None, searchArea="on_top_of",
...
-                                                    'arrived': 'SEGMENT'})
+                                                    'arrived': 'RISE'})
             else:
                 smach.StateMachine.add('NAVIGATE_TO_INSPECT', NavigateToObserve(robot, entityDes, radius=1.0),
                                        transitions={'unreachable': 'failed',
                                                     'goal_not_defined': 'failed',
-                                                    'arrived': 'SEGMENT'})
+                                                    'arrived': 'RISE'})
+
+            smach.StateMachine.add('RISE', RiseForInspect(robot, entityDes, searchArea),
```
jlunenburg: We might consider segmenting even if rising failed. Segmentation might still work, won't it?

robot_smach_states/src/robot_smach_states/rise.py
```diff
@@ -2,69 +2,67 @@
...
+    State to ensure that the robot is in a proper position to inspect. This means its head is at the correct height
+    and his vision is unobstructed. (no arm in front of its face)
     :param robot: Robot to execute state with
+    :param entity: entity which is to be inspected
+    :param volume: volume of the entity which is to be inspected
     """
-    def __init__(self, robot):
-        smach.State.__init__(self, outcomes=['succeeded', 'failed'])
+    def __init__(self, robot, entity, volume=None):
+        smach.State.__init__(self, outcomes=['done', 'failed'])
```
jlunenburg: Why change this from 'succeeded' to 'done'? 'Done' might also imply that it has failed so I actually prefer 'succeeded'.

robot_smach_states/src/robot_smach_states/world_model.py
```diff
@@ -203,12 +204,16 @@ def __init__(self, robot, entityDes, objectIDsDes=None, searchArea="on_top_of",
                                                                                  entityDes),
                                        transitions={'unreachable': 'failed',
                                                     'goal_not_defined': 'failed',
-                                                    'arrived': 'SEGMENT'})
+                                                    'arrived': 'RISE'})
```
jlunenburg: There's no mechanism that ensures that the entity designator always resolves to the same entity (was already the case in this state). IMHO we should start thinking about this so that we *either* assert that is a locking designator and lock it (not so convenient for the user) *or* use userdata to this end (easier for the end-user, less reusable).

robot_skills/src/robot_skills/hero.py
```diff
@@ -50,6 +50,63 @@ def __init__(self, wait_services=False):
...
+        """
+        # calculate the arm_lift_link which must be sent
+        z_over = 0.3  # height the robot should look over the surface
+        z_hh = 0.9  # height of the robots head at z_arm=0
+        torso_to_arm_ratio = 2  # motion of arm/motion of torso
+        z_head = inspect_target.z() + z_over
+
+        # check whether moving the arm is necessary
+        if z_head < 1.2:
+            rospy.loginfo('Entity is low enough. we dont need to move the arm')
```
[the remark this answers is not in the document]
PetervDooren: When should I use logdebug vs loginfo vs logwarn?

robot_smach_states/src/robot_smach_states/world_model.py
jlunenburg: Did you address this?

PetervDooren: I did not address this. Does this still fall in the scope of this PR? I think that locking the designator is the safest option however I fear this will break many applications.

PR #861, opened 2019-06-12
Add robotics_testlabs where_is_this knowledge

MatthijsBurgh:


robocup_knowledge/src/robocup_knowledge/environments/robotics_testlabs/challenge_where_is_this.py
```diff
@@ -0,0 +1,48 @@
+# RIPS KNOWLEDGE FILE RGO2019
```
JosjaG: If you use comments, at least make sure they are correct 😆

PR #872, opened 2019-07-04
Fix/remove start errors

jlunenburg:


robot_skills/src/robot_skills/hero.py
```diff
@@ -1,4 +1,5 @@
 from robot_skills import robot, api, arms, base, ebutton, head, ears, lights, perception, speech, torso, world_model_ed
+from .simulation import is_sim_mode, SimEButton
```
MatthijsBurgh: Please add to the import from robot_skills

PR #876, opened 2019-07-10
feat(smach)Added NavigateRobust

jlunenburg:
This is what I had in mind for NavigateRobust


LoyVanBeek:
This is actually kind-of like a Fallback aka Selector in a behavior tree.

robot_smach_states/src/robot_smach_states/navigation/navigate_robust.py
```diff
@@ -0,0 +1,70 @@
...
+                # Add an option
+                smach.StateMachine.add("NAVIGATE{}".format(idx), option,
+                                       transitions={
+                                           "arrived": "arrived",
+                                           "unreachable": "SAY{}".format(idx),
+                                           "goal_not_defined": "SAY{}".format(idx),
+                                       })
+
+                # Have the robot say something
+                smach.StateMachine.add("SAY{}".format(idx),
```
MatthijsBurgh: Say already randomly selects sentences from a list, so no need to rotate ourselves IMO.

jlunenburg: > Say already randomly selects sentences from a list, so no need to rotate ourselves IMO.

Depends on whether you want this deterministic or random. I opted for the determinism but feel free to challenge this

LoyVanBeek: Instead of the determinism, I'd make saying anything at all optional, also for the normal nav states.

jlunenburg: Out of scope ;-)

> Instead of the determinism, I'd make saying anything at all optional, also for the normal nav states.

Out of scope ;-)

LarsJanssenTUe:

PR #886, opened 2019-08-24
StartChallengeRobust wait for local planner footprint

### LINKED ISSUE #785: Fix challenge-robust

MatthijsBurgh:
Instead of waiting for 5 seconds, wait till the footprint of the local planner is published.

fixed #785


robot_skills/src/robot_skills/base.py
```diff
@@ -134,7 +134,7 @@ def getPlan(self, position_constraint):
     def checkPlan(self, plan):
         try:
             resp = self._check_plan_client(plan)
-        except:
+        except Exception:
             rospy.logerr("Could not check plan, is the global planner running?")
```
LoyVanBeek: Please also print the error, especially since this is a very broad exception class.

If possible, also catch a more narrow exception

MatthijsBurgh: Done

robot_skills/src/robot_skills/base.py
```diff
@@ -134,8 +134,9 @@ def getPlan(self, position_constraint):
     def checkPlan(self, plan):
         try:
             resp = self._check_plan_client(plan)
-        except:
+        except Exception as e:
             rospy.logerr("Could not check plan, is the global planner running?")
+            rospy.logerr(e)
```
jlunenburg: Why don't you do it in a single log statement:
```python
rospy.logerr("Could not check plan: {}".format(e.message))
```
That would result in a sinlge logrecord of an exception instead of multiple ones. This is much clearer in e.g., rqt-console (and even the terminal: less clutter by timestamps and levels)

robot_skills/src/robot_skills/world_model_ed.py
```diff
@@ -123,8 +123,9 @@ def get_closest_entity(self, type="", center_point=None, radius=0):
         try:
             center_in_map = center_point.projectToFrame("/map", self._tf_listener)
             entities = sorted(entities, key=lambda entity: entity.distance_to_2d(center_in_map.vector))
-        except:
+        except Exception as e:
             rospy.logerr("Failed to sort entities")
+            rospy.logerr(e)
```
jlunenburg: Idem

robot_skills/src/robot_skills/world_model_ed.py
```diff
@@ -178,8 +179,9 @@ def get_closest_laser_entity(self, type="", center_point=VectorStamped(), radius
             entities = sorted(entities, key=lambda entity: entity.distance_to_2d(
                 center_point.projectToFrame("/%s/base_link" % self.robot_name,
                                             self._tf_listener).vector))  # TODO: adjust for robot
-        except:
+        except Exception as e:
             print("Failed to sort entities")
+            rospy.logerr(e)
```
jlunenburg: While you're at it, please replace this print statement by a proper logmessage as well

robot_skills/src/robot_skills/world_model_ed.py
```diff
@@ -318,8 +318,9 @@ def get_closest_possible_person_entity(self, center_point=VectorStamped(), radiu
         try:
             entities = sorted(entities, key=lambda entity: entity.distance_to_2d(center_point.vector))
             rospy.logdebug("entities sorted closest to robot = {}".format(entities))
-        except:
+        except Exception as e:
```
jlunenburg: Idem

robot_smach_states/src/robot_smach_states/startup.py
```diff
@@ -239,3 +239,44 @@ def execute(self, userdata=None):
 
         rospy.loginfo("Timed out with door still closed")
         return "closed"
+
+
+class WaitForLocalPlanner(smach.State):
+    def __init__(self, robot, timeout):
```
jlunenburg: Docstrings missing ;-). Especially important for robot skills and robot smach states: if someone wants to use it, he/she shouldn't have to look at the code

robot_smach_states/src/robot_smach_states/startup.py
```diff
@@ -239,3 +239,44 @@ def execute(self, userdata=None):
...
+        self.footprint_sub = None
+
+    def msg_cb(self, msg):
+        # type: (PolygonStamped) -> None
+        try:
+            if len(msg.polygon.points) >= 3:
+                rospy.loginfo("Valid footprint received")
+                self._local_planner_ready.set()
+        except Exception as e:
+            rospy.logerr("Failed to receive footprint, so unsubscribing")
```
jlunenburg: Same logging comment as before

robot_skills/src/robot_skills/perception.py
```diff
@@ -267,7 +271,7 @@ def get_face_properties(self, faces=None, image=None):
             face_properties_response = self._face_properties_srv(faces)
             face_properties = face_properties_response.properties_array
         except Exception as e:
-            rospy.logerr(str(e))
+            rospy.logerr(e)
```
jlunenburg: You're sure this works and doesn't expect a string?

robot_skills/src/robot_skills/world_model_ed.py
```diff
@@ -195,7 +193,11 @@ def get_entity(self, id, parse=True):
         return entities[0]
 
     def get_entity_info(self, id):
-        return self._ed_entity_info_query_srv(id=id, measurement_image_border=20)
+        try:
```
jlunenburg: Might be out of scope but in general I actually don't like catching all exceptions at this level: it implies you always have to check all return values (is it empty? is it 'None'?). And if so, what do I return if to the caller above? IMHO, exceptions should be caught at the level that they can be handled appropriately.

robot_smach_states/src/robot_smach_states/startup.py
```diff
@@ -239,3 +239,59 @@ def execute(self, userdata=None):
...
+        :param robot: robot object
+        :type robot: robot
+        :param timeout: timeout
+        :type timeout: (float, int)
+        """
+        smach.State.__init__(self, outcomes=["ready", "timeout"])
+        self._robot = robot
+        self._timeout = timeout
+
+        self._local_planner_ready = Event()
```
jlunenburg: In my previous review I left a commented that didn't come through: I think you can get rid of the local planner ready and footprint sub member attributes. Concerning the subscriber: you also have to unregister upon fail, so why do it in the callback as well?

The event is a bit more challenging but could also work:
```python
def msg_cb(self, ready_event, msg):
    try:
        if len(msg.polygon.points) >= 3:
            rospy.loginfo("Valid footprint received")
            ready_event.set()
    except Exception as e:
        pass

def execute(self, userdate=None):
    rospy.loginfo("Waiting for local planner footprint")
    footprint_topic = "/{}/local_planner/local_costmap/robot_footprint/footprint_stamped".\
        format(self._robot.robot_name)

    ready_event = Event()
    footprint_sub = rospy.Subscriber(footprint_topic, PolygonStamped, functools.partial(self.msg_cb, ready_event))

    ready_before_timeout = self._local_planner_ready.wait(self._timeout)

    rospy.loginfo("Unregistering footprint subscriber")
    footprint_sub.unregister()

    if ready_before_timeout:
        rospy.loginfo("Local Planner is ready")
        return "ready"

    rospy.loginfo("Local Planner not ready during timeout")
    return "timeout"
```
Check out, e.g., https://answers.ros.org/question/39774/subscriber-callback-function-with-multiple-parameters/

Finally, I don't see why the try-except block is necessary

robot_skills/src/robot_skills/world_model_ed.py
MatthijsBurgh: True, but I found that all service calls in this file were wrapped in a try/except. Therefore I also wrapped this service call.

robot_skills/src/robot_skills/perception.py
MatthijsBurgh: Yep, Python resolves this automatically to the string representation of the Exception.

PR #901, opened 2019-09-24
Rwc2019 challenge hand me that

JosjaG:


challenge_hand_me_that/src/challenge_hand_me_that/get_furniture_from_operator_pose.py
```diff
@@ -0,0 +1,175 @@
...
+                          'dishwasher',
+                          'desk',
+                          'coffee_table',
+                          # 'couch',
+                          # 'armchair',
+                          'display_cabinet',
+                          'sideboard']
+
+
+class GetFurnitureFromOperatorPose(StateMachine):
```
JosjaG: This state, if cleaned up, could be a nice addition to robot_smach_states.

MatthijsBurgh: Cleanup is an understatement

JosjaG:
GetFurnitureFromOperator has been made into a robot_smach_state, which has a separate pull request. So IMO this can be merged.

PR #907, opened 2019-10-04
Split doctests

MatthijsBurgh:
Merge together with tue-robotics/tue-env-targets#84


robot_smach_states/test/test_doctest.py
```diff
@@ -1,37 +1,15 @@
...
-        print("Attempted count: {}".format(attempted_count))
-        self.assertEqual(failed_count, 0, "{} out of {} tests failed".format(failed_count, attempted_count))
+class DocTestsRobotSmachStates(_TestDocTests):
+    def __init__(self, method_name="test_doctests"):
+        super(DocTestsRobotSmachStates, self).__init__(module_name="robot_smach_states", method_name=method_name)
 
 
 if __name__ == '__main__':
-    unittest.main()
+    suite = unittest.TestSuite()
```
jlunenburg: What does this change do?

test_tools/src/test_tools/doctest_unittest.py
```diff
@@ -0,0 +1,36 @@
+#! /usr/bin/env python
+
+import doctest
+import importlib
+import os
+import rospkg
+import unittest
+
+
+class _TestDocTests(unittest.TestCase):
```
jlunenburg: I don't think this should be private if you intend to import this from other files/modules/packages which you do). Why did you choose to do so?

test_tools/src/test_tools/doctest_unittest.py
```diff
@@ -0,0 +1,36 @@
...
+
+    def __init__(self, module_name, method_name="test_doctests"):
+        super(_TestDocTests, self).__init__(method_name)
```
jlunenburg: Add docstring:
* `module_name`: is this only, e.g., `robot_smach_states`?
* what does the `method_name` do?

test_tools/src/test_tools/doctest_unittest.py
MatthijsBurgh: To prevent nosetest from running this class in the test files which import this class

robot_smach_states/test/test_doctest.py
MatthijsBurgh: This might not be needed anymore.

test_tools/src/test_tools/doctest_unittest.py
MatthijsBurgh: The method name is the name of the member function to run.

PR #909, opened 2019-10-08
test/challenge demo

jlunenburg:


challenge_demo/test/grammar_test.py
```diff
@@ -0,0 +1,130 @@
...
+        # In principle, a list of actions can be provided to the actions server
+        # Therefore, we need to take to steps to list all actions
+        # I.e., the following actions_list_target refers to the grammar target
+        # containing a list of actions
+        actions_list_target = ""
+        for option in root_rule.options:
+            if "actions" not in option.lsemantic:
+                continue
+            actions_list_target = option.conjuncts[0].name  # This is too hardcoded
+            break
```
LarsJanssenTUe: As far as I see it the action_list_target is only a single string right? So either it is not a list, or the list is within the string?

challenge_demo/test/grammar_test.py
```diff
@@ -0,0 +1,130 @@
...
+            "Don't know how to continue if the actions list rule has multiple options"
+        assert len(actions_list_rule.options[0].conjuncts) == 1, \
+            "Don't know how to continue if the actions options list rule has multiple conjunctions"
+
+        # Now, we need to get all separate options
+        actions_target = actions_list_rule.options[0].conjuncts[0].name
+
+        cls.actions_rule = cls.parser.rules[actions_target]
+
+    def test_grammar(self):
```
LarsJanssenTUe: docstring

challenge_demo/test/grammar_test.py
```diff
@@ -0,0 +1,130 @@
...
+from action_server.actions.action import ConfigurationResult
+from action_server.task_manager import TaskManager
+
+TMTestResult = collections.namedtuple("TMTestResult", ["recipe", "config_result"])
+
+
+class GrammarTest(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
```
LarsJanssenTUe: docstring

challenge_demo/test/grammar_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+        # Now, we need to get all separate options
+        actions_target = actions_list_rule.options[0].conjuncts[0].name
+
+        cls.actions_rule = cls.parser.rules[actions_target]
+
+    def test_grammar(self):
+
+        test_results = {}
+        for option in self.actions_rule.options:
```
LarsJanssenTUe: it is assumed that self.actions_rule.options is none empty. Should that be checked to be able to give a warning?

challenge_demo/test/grammar_test.py
```diff
@@ -0,0 +1,130 @@
...
+
+        :param target: target to look for
+        :return: option string
+        """
+        if target not in cls.parser.rules:
+            return target + " "
+
+        result = ""
+        rule = cls.parser.rules[target]
+        # For now, take the first available option
```
LarsJanssenTUe: What should this become?

challenge_demo/test/grammar_test.py
```diff
@@ -0,0 +1,130 @@
...
+                )  # type: ConfigurationResult
+            except Exception as e:
+                test_result = ConfigurationResult()
+                test_result.message = "Configuration crashed: {}".format(e.message)
+            test_results[result_str] = TMTestResult(actions_definition["actions"], test_result)
+
+        failed_test_results = {}
+        for action, test_result in test_results.iteritems():
+            if test_result.config_result.succeeded:
+                print("Configuration of '{}' succeeded".format(action))
```
LarsJanssenTUe: Why prints?

challenge_demo/test/grammar_test.py
```diff
@@ -0,0 +1,130 @@
+from __future__ import print_function
```
LarsJanssenTUe: TBH: the logic of the code looks good to me, however I do not completely understand the working

challenge_demo/test/grammar_test.py
jlunenburg: It's the 'target', i.e., the string identifying the underlying list of actions in the grammar

challenge_demo/test/grammar_test.py
jlunenburg: It's possible to iterate over an empty list. E.g.,
```python
for item in []:
    print(item)
```
doesn't crash. However, this would mean that the code doesn't test anything...

challenge_demo/test/grammar_test.py
jlunenburg: What else? If I use rospy I need a rostest instead of a pytest. This is extremely inconvenient. I don't want print statements in my library code but in tests I do find it acceptable.

challenge_demo/test/grammar_test.py
jlunenburg: See 74976f4b5effd72b089bd6e1d333fc2a0dc98fc9

challenge_demo/test/grammar_test.py
jlunenburg: See 74976f4b5effd72b089bd6e1d333fc2a0dc98fc9

challenge_demo/test/grammar_test.py
jlunenburg: See 74976f4b5effd72b089bd6e1d333fc2a0dc98fc9

challenge_demo/test/grammar_test.py
jlunenburg: I added a check in eb78a9547f16cc368838786a8024ebe356a18b89 that tests if this test actually does something.

PR #913, opened 2019-10-15
test(demo)Rewrote test to use action server test tool

jlunenburg:
N.B., this depends on https://github.com/tue-robotics/action_server/pull/120


challenge_demo/test/grammar_test.py
```diff
@@ -1,148 +1,11 @@
-from __future__ import print_function
-import collections
 import unittest
+from action_server.test_tools import test_grammar
```
MatthijsBurgh: Add test dependency of `action_server` to package.xml

challenge_demo/test/grammar_test.py
```diff
@@ -1,148 +1,11 @@
...
 
-from grammar_parser.cfgparser import CFGParser, Conjunct, Option, Rule
 from robocup_knowledge import load_knowledge
```
MatthijsBurgh: Add test dependency to `robocup_knowledge` to package.xml

PR #918, opened 2019-10-22
Feature(base destructor): add safety; robot stops driving on destruct…

LarsJanssenTUe:
…ion of base

### DIFF
--- a/robot_skills/src/robot_skills/base.py
+++ b/robot_skills/src/robot_skills/base.py
@@ -290,6 +290,13 @@ class Base(RobotPart):
         rospy.logwarn("[base.py] Function 'cancel_goal' of 'Base' is obsolete.")
         return True
 
+    def __del__(self):
+        v = geometry_msgs.msg.Twist()  # Initialize velocity
+        # Stop driving
+        v.linear.x = 0.0
+        v.linear.y = 0.0
+        v.angular.z = 0.0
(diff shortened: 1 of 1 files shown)


LarsJanssenTUe:
> This mean that when you close a console while a challenge is running, the robot stops moving.
> Please think sbout the potential side-effects really well

TBH: Then that should just not happen right? This could happen when there are no clear rules on how to use the terminal in challenges, which in my opinion is a solvabele problem. On the other side if, for example a control to pose is used and something crashes the robot will just continue driving, which is a genuine safety issue. 

@MatthijsBurgh what do you think?

MatthijsBurgh:
As also discussed with @jlunenburg, the idea behind this is correct. But in reality, `rospy` probably dies before the base is deleted. In that case this even doesn't work. Therefore, combined with possible unwanted side effects, we shouldn't implement this and just press that emergency button.

jlunenburg:
You could (try to) do this completely different:
* only publishing a zero velocity doesn't help: if the local planner is still running it will keep on going anyway so you need to take proper care of cancelling everything (although this is generally undesired but might be convenient in some cases)
* you could try to, e.g., catch the SIGINT [signal](https://docs.python.org/3/library/signal.html), cancel everything and *only then* call a `rospy.signal_shutdown()`. This would need some proper testing though

Issue #767, opened 2019-04-25
Hacks in get_arm API


MatthijsBurgh:
Introduced in cf87033ffe81b378eb1def10acafc3d952f86ffc

The whole arms stuff require some serious thoughts about the design.

alberth:
You want the challenge to work => you need to verify beforehand that the robot can do everything you want => you need a decision procedure for that.

Specifying in a challenge what you expect from an arm implements the latter for an arm.

So what is wrong with it?

PR #923, opened 2019-11-02
Enforce user request about accessing the force sensor

alberth:
Make it impossible to access the force sensor if the user din't ask for it, even if the hardware supports it.

### DIFF
--- a/robot_skills/src/robot_skills/arms.py
+++ b/robot_skills/src/robot_skills/arms.py
@@ -44,6 +44,9 @@ class PublicArm(object):
     """
     Public arm interface, also checks a challenge doesn't try to use more than it asked for.
 
+    THE ARM SHOULD NEVER ALLOW ANYTHING THAT WAS NOT ASKED FOR IN THE CONSTRUCTOR!
+
+
     :ivar _arm: Private link to the real arm.
     :vartype _arm: Arm
 
@@ -56,6 +59,9 @@ class PublicArm(object):
     :ivar _has_occupied_by: Whether the arm supports 'occupied_by' calls.
     :vartype _has_occupied_by: bool
 
+    :ivar _allow_force_sensor: Whether use of the force sensor is allowed.
+    :vartype _allow_force_sensor: bool
+
     :ivar _available_joint_goals: Joint goals that may be used.
     :vartype _available_joint_goals: set of str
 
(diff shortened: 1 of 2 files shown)


MatthijsBurgh:
If we are going to enforce this, it should be made more generic for different kind of sensors.

alberth:
It should become more generic indeed.

I considered that, but the pending issue is that there is currently no agreement on how to check in advance that a challenge will not fail on lack of hardware capabilities. For this reason, it seems unwise to spend time on generalizing code that may get deleted or moved to elsewhere. I started a discussion with a question about this in #767 .

This patch is just a fix wrt the semantics of `get_arm`. If you don't ask for the sensor there, you get an arm that does allow usage (if the hardware supports it). That is breaking the `get_arm` contract.

LarsJanssenTUe:

alberth:
No idea what happened, rebase just worked, just needed a force-push.

Obviously, this change is enforcing getting access to arm functionality by first querying about it, which seems still undecided.
Besides that, just merge, when the arm is refactored, this sensor thing is a simple thing.

robot_skills/examples/example_move_down_until_force_sensor_edge_up.py
```diff
@@ -4,6 +4,6 @@
 
 rospy.init_node('test_mode_down_until_force_sensor_edge_up', anonymous=True)
 robot = robot_constructor("hero")
-arm = robot.get_arm()
+arm = robot.get_arm(available_force_sensor=True)
```
LarsJanssenTUe: Make in line with master (force_sensor_required=True)

robot_skills/src/robot_skills/arms.py
```diff
@@ -167,11 +175,14 @@ def send_gripper_goal(self, state, timeout=5.0, gripper_type=None, max_torque=0.
 
     @property
     def has_force_sensor(self):
+        # Check that the user enabled force sensor access.
+        self._test_die(self._allow_force_sensor, 'allow_force_sensor=' + str(self._allow_force_sensor),
+                       "Specify get_arm(..., available_force_sensor=True)")
```
LarsJanssenTUe: available_force_sensor --> force_sensor_required

robot_skills/src/robot_skills/arms.py
```diff
@@ -167,11 +175,14 @@ def send_gripper_goal(self, state, timeout=5.0, gripper_type=None, max_torque=0.
...
         return hasattr(self._arm, "force_sensor")
 
     def move_down_until_force_sensor_edge_up(self, timeout=10, retract_distance=0.01):
-        self._test_die(self.has_force_sensor, 'available_force_sensor=' + str(self.has_force_sensor),
-                       "Specify get_arm(..., available_force_sensor=True")
+        self._test_die(self.has_force_sensor, 'has_force_sensor=' + str(self.has_force_sensor),
+                       "Specify get_arm(..., available_force_sensor=True)")
```
LarsJanssenTUe: available_force_sensor --> force_sensor_required

Issue #873, opened 2019-07-04
Fix designator.write([None]) exception


LoyVanBeek:
```
None [<class 'robot_skills.util.entity.Entity'>]
[ERROR] [1562215087.757899, 747.813000]: InvalidUserCodeError: Could not execute state 'FIND_PEOPLE' of type '<robot_smach_states.human_interaction.find_people_in_room.FindPeople object at 0x7f5d44c8a210>': Traceback (most recent call last):
File "/opt/ros/kinetic/lib/python2.7/dist-packages/smach/state_machine.py", line 247, in _update_once
self._remappings[self._current_label]))
File "/home/rein/dev/tue/src/tue_robocup/robot_smach_states/src/robot_smach_states/human_interaction/find_people_in_room.py", line 187, in execute
self._found_people_designator.write(result_people)
File "/home/rein/dev/tue/src/tue_robocup/robot_smach_states/src/robot_smach_states/util/designators/core.py", line 241, in write
elif isinstance(value, self.variable_designator.resolve_type):
TypeError: isinstance() arg 2 must be a class, type, or tuple of classes and types

[ERROR] [1562215087.758699, 747.816000]: State 'WAIT_FOR_CUSTOMER' failed to execute.
[ERROR] [1562215087.759029, 747.816000]: Robot exited with <class 'smach.exceptions.InvalidUserCodeError'>,Could not execute state 'FIND_PEOPLE' of type '<robot_smach_states.human_interaction.find_people_in_room.FindPeople object at 0x7f5d44c8a210>': Traceback (most recent call last):
File "/opt/ros/kinetic/lib/python2.7/dist-packages/smach/state_machine.py", line 247, in _update_once
self._remappings[self._current_label]))
File "/home/rein/dev/tue/src/tue_robocup/robot_smach_states/src/robot_smach_states/human_interaction/find_people_in_room.py", line 187, in execute
self._found_people_designator.write(result_people)
File "/home/rein/dev/tue/src/tue_robocup/robot_smach_states/src/robot_smach_states/util/designators/core.py", line 241, in write
elif isinstance(value, self.variable_designator.resolve_type):
TypeError: isinstance() arg 2 must be a class, type, or tuple of classes and types
,<traceback object at 0x7f5d44bed368>
[INFO] [1562215087.759412, 747.816000]: left arm cancelling all goals on all arm-related ACs on close
[INFO] [1562215087.759960, 747.816000]: Torso cancelling all goals on close
[...]

PR #924, opened 2019-11-02
Fix handling of designator None types/values

alberth:
Should address #873 but not sure. @LoyVanBeek ?


robot_smach_states/src/robot_smach_states/util/designators/core.py
```diff
@@ -27,25 +27,41 @@ def __init__(self, initial_value=None, resolve_type=None, name=None):
...
+                    raise TypeError("The resolve_type may not be [None]")
+            self._resolve_type = resolve_type
+        else:
+            # No resolve type provided, use the inital value to derive a type.
+            if self.__initial_value is None:
+                raise TypeError("resolve_type could not be inferred from None.")
+
             if isinstance(self.__initial_value, list):
-                if self.__initial_value:
+                if self.__initial_value: # Not an empty list.
```
MatthijsBurgh:
```suggestion
                if len(self.__initial_value) > 0:
```

robot_smach_states/src/robot_smach_states/util/designators/core.py
```diff
@@ -55,20 +71,32 @@ def resolve(self):
...
-            result_type = type(result[0])
+        if result is None: # No data available for checking the type.
+            return result
+
+        if isinstance(self._resolve_type, list):
+            # Not None, list expected instead.
+            if not isinstance(result, list):
+                self.fail_with_type_error(result_type, resolve_type, result)
+
+            if not result: # Empty list, cannot check element type.
```
MatthijsBurgh:
```suggestion
            if len(result) == 0: # Empty list, cannot check element type.
```

robot_smach_states/src/robot_smach_states/util/designators/core.py
```diff
@@ -55,20 +71,32 @@ def resolve(self):
...
 
-        if result is not None and not issubclass(result_type, resolve_type):
-            raise TypeError("{} resolved to a '{}' instead of expected '{}'. "
-                            "Originals: result type: {}, resolve type: {}".format(self, result_type, resolve_type,
-                                                                                  type(result), self.resolve_type))
         return result
 
+    def fail_with_type_error(self, result_type, resolve_type, result):
+            msg = "{} resolved to a '{}' instead of expected '{}'. Originals: result type: {}, resolve type: {}"
+            raise TypeError(msg.format(self, result_type, resolve_type, type(result), self.resolve_type))
```
MatthijsBurgh: resolve_type = self.resolve_type = property(_get_resolve_type) = return self._resolve_type

I think resolve type is now printed two times, both time provided via a completely different route. This should be simplified.

Both calls to `fail_with_type_error` have the same arguments, so t

robot_smach_states/src/robot_smach_states/util/designators/core.py
```diff
@@ -316,6 +338,55 @@ def test_list(self):
         self.assertEqual(v.resolve_type, [str])
         self.assertListEqual(v.resolve(), ['a', 'b', 'c'])
 
+    def test_none_value(self):
```
MatthijsBurgh: These test are moved to the test folder in `rwc2019_robot_smach_states`, this PR should hold till that PR is merged and then these changes should be moved to the test_folder as well.

robot_smach_states/src/robot_smach_states/util/designators/core.py
alberth: Some text seems to be missing at the end?

I moved some `[0]` index operations it seems, which makes it print duplicate information more often than it did in the past. I'll look into it.

robot_smach_states/src/robot_smach_states/util/designators/core.py
```diff
@@ -26,25 +26,41 @@ def __init__(self, initial_value=None, resolve_type=None, name=None):
...
         super(Designator, self).__init__()
 
         self._name = name
-
         self.__initial_value = initial_value
-        if not resolve_type:
+        if resolve_type:
+            if isinstance(resolve_type, list):
+                if len(resolve_type) != 1:
+                    msg = "The resolve_type must be a sinlge-item list, found {}."
```
LoyVanBeek: typo

robot_smach_states/src/robot_smach_states/util/designators/core.py
```diff
@@ -26,25 +26,41 @@ def __init__(self, initial_value=None, resolve_type=None, name=None):
...
         """
         super(Designator, self).__init__()
 
         self._name = name
-
         self.__initial_value = initial_value
-        if not resolve_type:
+        if resolve_type:
+            if isinstance(resolve_type, list):
+                if len(resolve_type) != 1:
```
LoyVanBeek: I think there are cases where a list of types is used as 'either one of these types' but if so (and I'm not sure), then _that_ use-case is just wrong.

robot_smach_states/src/robot_smach_states/util/designators/core.py
```diff
@@ -130,17 +158,11 @@ def __init__(self, initial_value=None, resolve_type=None, name=None):
...
         super(VariableDesignator, self).__init__(initial_value, resolve_type, name=name)
-
         self._current = initial_value
-
         self.writeable = VariableWriter(self)
 
     def _set_current_protected(self, value):
         resolve_type = self.resolve_type
-        try:
-            resolve_type = tuple(self.resolve_type)
```
LoyVanBeek: WTF was this?

robot_smach_states/test/util/designators/test_core.py
```diff
@@ -46,6 +46,55 @@ def test_list(self):
...
+            def __init__(self):
+                super(MyVariableDesignator, self).__init__(resolve_type=[object])
+            def _resolve(self):
+                return None
+
+        v = MyVariableDesignator()
+        self.assertEquals(v.resolve(), None)
+
+    def test_empty_list_value(self):
+        # Resolve a [object] to None value
```
LoyVanBeek: Docstring doesn't match content

robot_smach_states/test/util/designators/test_core.py
```diff
@@ -46,6 +46,55 @@ def test_list(self):
         self.assertEqual(v.resolve_type, [str])
         self.assertListEqual(v.resolve(), ['a', 'b', 'c'])
 
+    def test_none_value(self):
```
LoyVanBeek: :heart_eyes: love all these tests!

robot_smach_states/src/robot_smach_states/util/designators/core.py
alberth: Likely caused by confusion about the semantics of "[<type>]" vs "<type>" in `resolve_type`, although I didn't check history.

Storing a non-type value in a variable that should contain a type is quite confusing, the proper solution would be to make 2 classes, one for expressing a value for a singular value, and one for expressing a value for a list of items.

robot_smach_states/src/robot_smach_states/util/designators/core.py
alberth: Clarified semantics of resolve_type somewhat

MatthijsBurgh: The current implementation allows the following: `[[float], [int], float, int]`. So multiple resolve types are allowed. If one of the resolve types is a list, that list should contain all of the same types, so the `len([float]) == 1`.

alberth: In my changes, a single type means a value of that type, and list with a single type means a list of values of that type.
This is not really nice, but Python lacks the concept "list of <type>" in its type system.

As Loy says and I agree, there should not be a set of allowed types. Inheritance is supported though as it uses `isinstance`. That handles `int` vs `float` already. Mixing single values and lists of values is non-Pythonic, you should know what you get without code to figure that out.

LoyVanBeek: Another option, if we want to allow multiple types, is to pass that asa `set()`, which is also a builtin type. 
But I do prefer to just allow 1 type to be specified. 

So, this change if fine AFAIK.

MatthijsBurgh: Right now a mix of lists and single items is allowed. Also different single items/items in the list are allowed. So Before changing this API, please check the entire `tue_robocup` for compatibility.

robot_smach_states/test/util/designators/test_core.py
```diff
@@ -46,6 +46,55 @@ def test_list(self):
...
+
+    def test_none_type(self):
+        with self.assertRaises(TypeError):
+            v = VariableDesignator(resolve_type=[None])
+
+    def test_multi_element_type(self):
+        with self.assertRaises(TypeError):
+            v = VariableDesignator(resolve_type=[str, bool])
+
+    def test_none_type(self):
```
LoyVanBeek: This test name is used 2x, also on line 87

robot_smach_states/src/robot_smach_states/util/designators/core.py
LoyVanBeek: Yes, I've been pondering to split the VariableDesignator into one for collections and one for singular values. But That's for later I'd say

robot_smach_states/src/robot_smach_states/util/designators/core.py
alberth: I checked the designators in tue_robocup, and found no cases of having multiple types in the constructor call. If my check was incomplete, at least it is not a commonly used feature.

PR #929, opened 2019-11-14
Remove sys.exit from knowledge loader

MatthijsBurgh:


LoyVanBeek:
I'm out of the loop, which issue is solved by this?

MatthijsBurgh:
A problem I experienced by construction tests in CI. As knowledge is sometimes loaded outside the `setup_statemachine` functions. The knowledge is loaded on import. As during the import `sys.exit` is called, the actual test is never reached. This was very difficult to debug. An raise would be a much nicer solution

TL; DR:
I don't think anywhere in the code `sys.exit` should be called, except in the `__main__`.

LoyVanBeek:
> I don't think anywhere in the code `sys.exit` should be called, except in the `__main__`
Agreed

PR #952, opened 2019-12-10
Initial version of point detect smach state

JosjaG:
Merge after https://github.com/tue-robotics/action_server/pull/133


robot_smach_states/src/robot_smach_states/human_interaction/entity_from_operator_pose.py
```diff
@@ -0,0 +1,118 @@
...
+        :param robot: Robot API object
+        :param knowledge: available knowledge to be used for the furniture
+        """
+        State.__init__(self, outcomes=['succeeded', 'failed'])
+        self._robot = robot
+        self.operator = None
+        self.all_possible_furniture = knowledge.location_names
+
+    def execute(self, userdata=None):
+        rospy.loginfo("I entered the execution state!")
```
PetervDooren: This statement is unneccessary

robot_smach_states/src/robot_smach_states/human_interaction/entity_from_operator_pose.py
```diff
@@ -0,0 +1,118 @@
...
+                i=result.entity_id, p=result.intersection_point.point))
+
+            if result.entity_id in self.all_possible_furniture:
+                final_result = result
+            else:
+                rospy.loginfo("{} is not furniture".format(result.entity_id))
+                self._robot.speech.speak("That's not furniture, you dummy.")
+                rospy.sleep(3)
+                self.operator = None
+                self._robot.get_arm().send_joint_goal("reset")
```
PetervDooren: Why do we need to reset the arm?

robot_smach_states/src/robot_smach_states/human_interaction/entity_from_operator_pose.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def _show_image(self, timeout=5.0):  # could also use /hero/openpose/result_image as rgb (same msg type)
+        rgb, depth, depth_info = self._robot.perception.get_rgb_depth_caminfo()
+        self._robot.hmi.show_image_from_msg(rgb, timeout)
+        return rgb, depth, depth_info
+
+    def _prep_operator(self):
+        self.operator = None
+
+        self._robot.ed.reset()
```
PetervDooren: I would not reset ed. That is far too destructive. Why do we reset it?

robot_smach_states/src/robot_smach_states/human_interaction/entity_from_operator_pose.py
```diff
@@ -0,0 +1,118 @@
...
+                person = persons[0]
+                if self._is_operator(person):
+                    self.operator = person
+
+        # robot.speech.speak("I see an operator at %.2f meter in front of me" % OPERATOR.position.z)
+        rospy.loginfo("I see an operator at %.2f meter in front of me" % self.operator.position.z)
+
+        return
+
+    def _is_operator(self, person):
```
PetervDooren: I would rename this to something that better reflects what the function does.
like '_verify_operator_pose'

robot_smach_states/src/robot_smach_states/human_interaction/entity_from_operator_pose.py
```diff
@@ -94,15 +90,15 @@ def _get_operator(self):
             if persons:
                 persons = sorted(persons, key=lambda x: x.position.z)
                 person = persons[0]
```
alberth: Sorting an entire list and then take the first element is an expensive form of finding the element with the minimum value.
A cheaper solution is `min(persons, key=lambda p: p.position.z)`
For more details, see https://docs.python.org/2.7/library/functions.html?highlight=min#min

JosjaG:
Two comments based on testing:
* It would be better if the robot would wait until it sees an operator (with a timeout), instead of simply fail if it does not immediately find one. 
* The distance to operator is not correct; perhaps check if the correct set of joints can be found and if not, based on the amount of joints that can be found ask the operator to move closer or further away.

robot_smach_states/src/robot_smach_states/human_interaction/entity_from_operator_pose.py
```diff
@@ -0,0 +1,114 @@
...
+            except:
+                return 'failed'
+                # failures should be caught
+            if persons:
+                persons = sorted(persons, key=lambda x: x.position.z)
+                person = persons[0]
+                if self._verify_operator_pose(person):
+                    self.operator = person
+
+        # robot.speech.speak("I see an operator at %.2f meter in front of me" % OPERATOR.position.z)
```
MatthijsBurgh:
```suggestion
```

JosjaG:
Another comment based on testing:
- There's a bug in the ray tracer (in the PointRayTraceResult (in raytrace.cpp)), will make a separate issue

JosjaG:
@MatthijsBurgh do you know why some of the checks fail?

MatthijsBurgh:
> @MatthijsBurgh do you know why some of the checks fail?

All separate issues

MatthijsBurgh:
apt-get was no installing recommended targets anymore. A PR, https://github.com/tue-robotics/tue-env/pull/469, to revert this has been opened. This caused some libraries to be missing.

I have removed the operator cleanup from the cleanup challenge, as this is not working as that ed plugin is not available. Which also removes the dependency on the Python Imaging Library, PIL, which was missing.

PR #956, opened 2019-12-14
Rwc2019 set the table

PetervDooren:


PetervDooren:
This pull request gives way to many conflicts. I cant find where it happens but it is too much effort for too little gain to find out. Hotfix for this is made in #959

PR #964, opened 2019-12-14
Use weak references to store active designators.

alberth:
Using a plain list makes created designators to exist forever. Using a weak reference to it allows it to be removed from memory, once nobody has a link to it.

As a result, the set will contain only the designators that are in use (or at least can be accessed).

Using a set means the order may get messed up, but a graph is built from the found designators, so the order doesn't seem highly relevant anyway.

### DIFF
--- a/robot_smach_states/src/robot_smach_states/util/designators/core.py
+++ b/robot_smach_states/src/robot_smach_states/util/designators/core.py
@@ -1,4 +1,5 @@
 __author__ = 'loy'
+import weakref
 from deprecation_warnings import get_caller_info
 import rospy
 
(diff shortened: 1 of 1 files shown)


LarsJanssenTUe:
Looks good to me! However I have not enough knowledge of the underlying working of designators to ensure that this will not break unexpectedly

LoyVanBeek:
> Looks good to me! However I have not enough knowledge of the underlying working of designators to ensure that this will not break unexpectedly

The Designator-class keeps a list of all designator-objects so we can do some analysis on them and the graph of them later on. 

But @LarsJanssenTUe just go through the core implementation and a few simple ones, it's rather simple and there is no magic to them :-)

PR #974, opened 2019-12-19
Faster empty spot designator

### LINKED ISSUE #943: Empty spot designator should be more efficient
For example take the first empty spot that you find

PetervDooren:
Fixes #943


robot_smach_states/src/robot_smach_states/util/designators/ed_designators.py
```diff
@@ -323,42 +323,36 @@ def _resolve(self):
         assert all(isinstance(v, FrameStamped) for v in vectors_of_interest)
 
         open_POIs = filter(self.is_poi_occupied, vectors_of_interest)
+        open_POIs.sort(key=self.distance_to_poi_area_heuristic)
 
-        return self.select_best_feasible_poi(open_POIs)
+        for poi in open_POIs:
+            if self.distance_to_poi_area(poi):
+                selection = self.create_selection_marker(poi)
```
LarsJanssenTUe: Is deze poi een "selected_pose" zoals de create_selection_marker def nodig heeft?

robot_smach_states/src/robot_smach_states/util/designators/ed_designators.py
```diff
@@ -323,42 +323,36 @@ def _resolve(self):
...
+        return None
 
     def is_poi_occupied(self, frame_stamped):
         entities_at_poi = self.robot.ed.get_entities(center_point=frame_stamped.extractVectorStamped(),
                                                      radius=self._spacing)
         return not any(entities_at_poi)
 
+    def distance_to_poi_area_heuristic(self, frame_stamped):
+        base_pose = self.robot.base.get_location()
+        bo = self.robot.arms.values()[0].base_offset
```
LarsJanssenTUe: I guess these querries will occur for each of the poi instances. Why not give these values as a parameter and thereby making the code even faster?

robot_smach_states/src/robot_smach_states/util/designators/ed_designators.py
```diff
@@ -323,42 +323,36 @@ def _resolve(self):
         assert all(isinstance(v, FrameStamped) for v in vectors_of_interest)
 
         open_POIs = filter(self.is_poi_occupied, vectors_of_interest)
+        open_POIs.sort(key=self.distance_to_poi_area_heuristic)
```
LarsJanssenTUe: There is no frame_stamped given to distance_to_poi_area_heuristic. Does this work? In case it does please specify how!

robot_smach_states/src/robot_smach_states/util/designators/ed_designators.py
PetervDooren: yes

PetervDooren:
The edge score is used to favor points in the middle of a long edge. To quote the comment "In case of a cabinet, e.g., this might prevent the robot from hitting the cabinet edges". I don't know if this is something that is still relevant. @jlunenburg added this 3 years ago.

While this new EmptySpotDesignator only takes the absolute length between the points. The EmptyspotDesignator is only used when placing. (It is too specific to be used for anything else) And in that case it is a fair assumption that we are already in front of the furniture on which to place something. Since a place is always preceded by an inspect.

PetervDooren:
As of #977 we no longer resolve the emptyspotdesignator three times every Place. I suggest we test on the real robot to determine whether the trade-off of speed vs optimality is worth it.

LarsJanssenTUe:
> The edge score is used to favor points in the middle of a long edge. To quote the comment "In case of a cabinet, e.g., this might prevent the robot from hitting the cabinet edges". I don't know if this is something that is still relevant. @jlunenburg added this 3 years ago.
> 
I think in itself is a good idea, but the question I ask myself is whether or not it is necessary. Again the question is performance vs speed, but by removing the fact that the emptyspotdesignator is resolved multiple times might already do a lot in terms of time. 

> While this new EmptySpotDesignator only takes the absolute length between the points. The EmptyspotDesignator is only used when placing. (It is too specific to be used for anything else) And in that case it is a fair assumption that we are already in front of the furniture on which to place something. Since a place is always preceded by an inspect.
>
I think we need to check whether the emptyspotdesignator is indeed not used elsewhere (this would be logical, but our code sometimes is not logical...). If this is the case you are quite right, under normal circumstances the distance from point to point would be fine. However if we have won enough in terms of speed I think I would prefer the length of the path for robustness reasons (right now I can't think of a way of how the point to point method would break, but that does not mean there is no such way!)

PetervDooren:
Just tested the old emptyspotdesignator. Just checking all the positions for the dinner table takes 2 seconds.

PetervDooren:
new version takes 0.7 seconds

robot_smach_states/src/robot_smach_states/util/designators/ed_designators.py
PetervDooren: I did this and it greatly improved speed

robot_smach_states/src/robot_smach_states/manipulation/place_designator.py
```diff
@@ -0,0 +1,247 @@
...
+        assert all(isinstance(v, FrameStamped) for v in vectors_of_interest)
+
+        open_POIs = filter(self.is_poi_occupied, vectors_of_interest)
+
+        base_pose = self.robot.base.get_location()
+        arm = self.arm_designator.resolve()
+        open_POIs_dist = [(poi, self.distance_to_poi_area_heuristic(poi, base_pose, arm)) for poi in open_POIs]
+
+        # We don't care about small differences
+        nav_threshold = 0.5 # Distance (0.5 m)
```
LarsJanssenTUe: Should we do something about this magic number? Maybe make it configurable with a default parameter?

robot_smach_states/src/robot_smach_states/manipulation/place_designator.py
```diff
@@ -0,0 +1,247 @@
...
+        open_POIs_dist = [f for f in open_POIs_dist if (f[1] - open_POIs_dist[0][1]) < nav_threshold]
```
LarsJanssenTUe:
```suggestion
        open_POIs_dist = open_POIs_dist.sort(key=lambda tup:tup[0])
        open_POIs_dist = [f for f in open_POIs_dist if (f[1] - open_POIs_dist[0][1]) < nav_threshold]
```

robot_smach_states/src/robot_smach_states/manipulation/place_designator.py
```diff
@@ -0,0 +1,247 @@
...
+
+        rospy.logerr("Could not find an empty spot")
+        return None
+
+    def is_poi_occupied(self, frame_stamped):
+        entities_at_poi = self.robot.ed.get_entities(center_point=frame_stamped.extractVectorStamped(),
+                                                     radius=self._spacing)
+        return not any(entities_at_poi)
+
+    def distance_to_poi_area_heuristic(self, frame_stamped, base_pose, arm):
```
LarsJanssenTUe: docstring

robot_smach_states/src/robot_smach_states/manipulation/place_designator.py
```diff
@@ -0,0 +1,247 @@
...
+        offset_pose_x = offset_pose[0]
+        offset_pose_y = offset_pose[1]
+
+        x = frame_stamped.frame.p.x()
+        y = frame_stamped.frame.p.y()
+
+        dist = math.hypot(offset_pose_x - x, offset_pose_y - y)
+        return dist
+
+    def distance_to_poi_area(self, frame_stamped, arm):
```
LarsJanssenTUe: docstring

LarsJanssenTUe:
@PetervDooren I forgot to test the added sort yesterday so this has to be done (preferably on the robot with timing) before this is merged... Sorry, my bad!

PetervDooren:
Tested on the robot. Time to resolve is around 0.7 seconds

LarsJanssenTUe:
> Tested on the robot. Time to resolve is around 0.7 seconds

Very nice! I like these kinds of upgrades in time!

PR #977, opened 2019-12-25
Fix/remove redundant resolving in the place smach state

PetervDooren:


LoyVanBeek:
Then, what determines at what pose to put an item? That is what the placement_pose is for.

PetervDooren:
It is still placement pose. But by putting in a locking designator it will only need to compute the optimal empty spot once. This also ensures that the pose that is navigated to is the pose that we put the object

robot_smach_states/src/robot_smach_states/manipulation/place.py
```diff
@@ -8,12 +8,14 @@
...
 from robot_smach_states.world_model import Inspect
+from robot_smach_states.utility import LockDesignator
 from robot_smach_states.util.designators.ed_designators import EmptySpotDesignator
+from robot_smach_states.util.designators.utility import LockingDesignator
 from robot_smach_states.util.designators import check_type
 
 
 class PreparePlace(smach.State):
-    def __init__(self, robot, placement_pose, arm):
+    def __init__(self, robot, arm):
```
MatthijsBurgh: Remove also from docstring

robot_smach_states/src/robot_smach_states/manipulation/place.py
```diff
@@ -252,16 +249,19 @@ def __init__(self, robot, item_to_place, place_pose, arm, place_volume=None, upd
...
                                        )
 
-            smach.StateMachine.add('PREPARE_PLACE', PreparePlace(robot, place_designator, arm),
-                                   transitions={'succeeded': 'NAVIGATE_TO_PLACE',
+            smach.StateMachine.add('PREPARE_PLACE', PreparePlace(robot, arm),
+                                   transitions={'succeeded': 'LOCK_DESIGNATOR',
                                                 'failed': 'failed'})
 
-            smach.StateMachine.add('NAVIGATE_TO_PLACE', NavigateToPlace(robot, place_designator, arm),
+            smach.StateMachine.add('LOCK_DESIGNATOR', LockDesignator(locking_place_designator),
```
MatthijsBurgh: Should we unlock it afterwards or not?

robot_smach_states/src/robot_smach_states/manipulation/place.py
PetervDooren: Done

robot_smach_states/src/robot_smach_states/manipulation/place.py
PetervDooren: We don't need to. The locking designator only exists within the scope of Place. After place has been executed we no longer need it.

PR #980, opened 2020-01-04
This reverts commit cf87033ffe81b378eb1def10acafc3d952f86ffc.

### LINKED ISSUE #767: Hacks in get_arm API

LarsJanssenTUe:
resolves #767


LarsJanssenTUe:
I know this will probably break a lot of things, but I think this is worth re-implementing. The question now is what the best way to proceed is (that is also why I made this pr).

alberth:
An issue existed that stated the check as a good direction, yet when it was implemented a hack was added hours after hitting master.
I would suggest to first decide what we actually want. It may save a lot of wasted time.

LarsJanssenTUe:
For a lot of things this seems to work out however it needs to be checked how this is gonna work for the Grab state, because things might go wrong here.....

challenge_set_a_table/src/challenge_set_a_table_states/manipulate_machine.py
```diff
@@ -130,7 +135,11 @@ def __init__(self, robot, place_designator):
     def execute(self, userdata=None):
         # Try to place the object
         item = ds.EdEntityDesignator(robot=self._robot, id=arm.occupied_by.id)
-        arm_designator = ds.OccupiedArmDesignator(self._robot)
+        arm_designator = ds.OccupiedArmDesignator(self._robot, arm_properties={"required_trajectories": "prepare_place",
+                                                                               "required_goals": ["reset",
+                                                                                                  "handover_to_human"],
```
alberth: Type of the `required_goals` parameter is not consistent. Here it is a list, at line 80 above, it is a string.
I quickly checked where the data goes, and it seems to be put into the `robot.get_arm` method. That method expects `a collection`, so a single string is likely not a collection that does anything useful.

This comment seems to apply at many other places in this patch.

alberth:
Perhaps there should be a class for storing wanted arm properties rather than this ad-hoc dictionary use. The advantage of the latter is of course its simplicity and being light-weight, but that is also its disadvantage.

No way to document what you can put into the structure and what values mean, no way to do any form of checking basic sanity until you try to construct an arm deep down in the robot rather than while you construct a properties object, no way to find all uses if you ever need to change the interface, no way to attach a method to it (and likely the list is not complete).

challenge_set_a_table/src/challenge_set_a_table_states/manipulate_machine.py
LarsJanssenTUe: This is solved!

LarsJanssenTUe:
> Perhaps there should be a class for storing wanted arm properties rather than this ad-hoc dictionary use. The advantage of the latter is of course its simplicity and being light-weight, but that is also its disadvantage.
> 
> No way to document what you can put into the structure and what values mean, no way to do any form of checking basic sanity until you try to construct an arm deep down in the robot rather than while you construct a properties object, no way to find all uses if you ever need to change the interface, no way to attach a method to it (and likely the list is not complete).

Issue #999

challenge_cleanup/src/challenge_cleanup/self_cleanup.py
```diff
@@ -261,20 +262,25 @@ def __init__(self, robot, selected_entity_designator, room_des):
...
-                                                "failed": "SAY_GRAB_FAILED"})
+            smach.StateMachine.add(
+                "GRAB",
+                robot_smach_states.Grab(
+                    robot,
+                    selected_entity_designator,
+                    ds.UnoccupiedArmDesignator(robot,
+                                               arm_properties={"required_trajectories": ["prepare_grasp"],
+                                                               "required_goals": ["carrying_pose"],
+                                                               "reequired_gripper_types": [arms.GripperTypes.GRASPING]},
```
alberth: typo

challenge_set_a_table/src/challenge_set_a_table_states/manipulate_machine.py
```diff
@@ -132,7 +137,12 @@ def __init__(self, robot, place_designator):
     def execute(self, userdata=None):
         # Try to place the object
         item = ds.EdEntityDesignator(robot=self._robot, id=arm.occupied_by.id)
-        arm_designator = ds.OccupiedArmDesignator(self._robot)
+        arm_designator = ds.OccupiedArmDesignator(self._robot, arm_properties={"required_trajectories": ["prepare_place"
+                                                                                                         ],
+                                                                               "required_goals": ["reset",
+                                                                                                  "handover_to_human"],
```
alberth: Maybe move the `arm_properties=` to the next line? that would give you a little extra space to cleanup the layout.

challenge_manipulation/src/manipulation.py
```diff
@@ -32,13 +31,11 @@
 from robot_smach_states.util.startup import startup
 from robot_smach_states import Grab
 from robot_smach_states import Place
-from robot_smach_states.util.geometry_helpers import *
```
alberth: +1

challenge_cleanup/src/challenge_cleanup/self_cleanup.py
```diff
@@ -331,21 +337,31 @@ def __init__(self, robot, selected_entity_designator, room_des):
...
+                                                            store_entity_des,
+                                                            ds.OccupiedArmDesignator(
+                                                                robot,
+                                                                arm_properties={
+                                                                    "required_trajectories": ["prepare_place"],
+                                                                    "required_gripper_types": [
+                                                                        arms.GripperTypes.GRASPING]},
+                                                                name="occupied_arm_designator"),
+                                                            "on_top_of"),
+                                   transitions={"done": "SAY_PLACE_SUCCESS", "failed": "SAY_PLACE_FAILED"})
```
alberth: The line is too long, and you break it too late, so you get long line at the top, and then everything pushed at the right side in very short sentences, with a large amount of unused space at the left side.

Breaking this up in smaller pieces makes it much nicer:
```
arm_properties={"required_trajectories": ["prepare_place"],
                           "required_gripper_types": [arms.GripperTypes.GRASPING]}
designator = ds.OccupiedArmDesignator(robot, arm_properties=arm_properties,
                                       name="occupied_arm_designator")
placement = robot_smach_states.Place(robot, selected_entity_designator, store_entity_des,
                                     designator, "on_top_of")
smach.StateMachine.add('PLACE_TO_STORE', placement,
                       transitions={"done": "SAY_PLACE_SUCCESS", "failed": "SAY_PLACE_FAILED"})
```

challenge_cleanup/src/challenge_cleanup/self_cleanup.py
```diff
@@ -336,30 +336,23 @@ def __init__(self, robot, selected_entity_designator, room_des):
                                    transitions={"done": "PLACE_IN_TRASH",
                                                 "failed": "SAY_PLACE_FAILED"})
 
+            arm_properties_place = {"required_trajectories": ["prepare_place"],
+                                    "required_gripper_types": [arms.GripperTypes.GRASPING]}
+            arm_designator_place = ds.OccupiedArmDesignator(robot, arm_properties_place, name="occupied_arm_designator")
```
alberth: Even dupliceren om unieke objecten te krijgen

PR #984, opened 2020-01-07
Feature/benchmark grasping

LoyVanBeek:
TODO:

- [x] nav to waypoint, inspect furniture and grab item
- [x] Turn around and drop item
- [x] Output a line to .csv file
- [x] Nicely say when you're going to drop an item
- [x] Navigate to inspect before actually inspecting
- [x] Optionally grab any object class (indicated by passing `*` for the class). In case the class is specified, record the class of the item we grasped, for use in a object recognition benchmark as well. 
- [x] Record the position of the object we grasped as well.
- [x] Iterate over a bunch of objects collectively
- [ ] <strike>Check success of grasping.</strike> Not possible yet, out of scope
- [x] Record timestamp to compare progress over time

Follow-up PR:
- [ ] Add analysis subcommand that outputs useful stats:
    - [x] avg, mean, stddev over inspection duration
    - [x] avg, mean, stddev over grasping duration
    - [ ] Stats on how often grasping succeeded overall and a distribution over height, eg. what % of grasps at z=0.1m succeeded vs grasps at z=0.8m
    - [ ] confusion matrix between recognized classes in case only 1 object is expected and recognized. Feed into object recognition analysis tools


LarsJanssenTUe:
ToDo: Add a navigate to inspect

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,105 @@
...
+    pass
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description="Measure the time it takes to grasp an entity when starting from a given pose. "
+                                                 "After the grasping is done, go back to the waypoint, "
+                                                 "turn around and drop the item, "
+                                                 "then turn around once more to be in the start configuration again")
+    parser.add_argument("cls", type=str, help="class of entity to grasp, eg. 'coke'")
+    parser.add_argument("support", type=str, help="ID of entity to grasp FROM ('grasp' entity is on-top-of this 'support' entity), eg. 'cabinet'")
```
jlunenburg: I'd say we need a statistically relevant test. So grab one item from the cabinet, kitchen table, desk, ... etc. (maybe up to 10) and compute that duration. Then we know if we really sped things up. I guess we need a default list of entities as well. Maybe even add it to the script: print something like "make sure you have placed entities on x, y, z and z, press enter to continue"

LoyVanBeek:
For the iterated case, the input configuration can be specified as a .csv file, specifying what class of objects should be on which supporting entity and start location.
The output is just the same .csv file with **also** a summary with statistics over the success, durations over the configuration.

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,184 @@
...
+                if len(matching_results) > 1:
+                    rospy.logwarn("There are multiple items OK to grab, will select the last one")
+                    record['observed_class'] = ' '.join([result.type for result in matching_results])
+                else:
+                    record['observed_class'] = matching_results[-1].type
+                selected_entity_id = matching_results[-1].id
+
+                rospy.loginfo("Selected entity {} for grasping".format(selected_entity_id))
+                grasp_entity = ds.EdEntityDesignator(robot, id=selected_entity_id)
+                record['id'] = selected_entity_id[:6]
```
LarsJanssenTUe: There is only one id right? So why take the first 6 numbers? Shouldn't this break?

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,184 @@
...
+
+                entity = grasp_entity.resolve()  # type: Entity
```
LarsJanssenTUe: This could be done without a designator right?

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,184 @@
...
+                if entity:
+                    vector_stamped = entity.pose.extractVectorStamped()  # type: VectorStamped
+                    record['x'] = '{:.3f}'.format(vector_stamped.vector.x())
```
LarsJanssenTUe: x_position_grasp_entity?

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,184 @@
...
+except ImportError:
+    pass
+
+def single_item(robot, results_writer, cls, support, waypoint, inspect_from_area=None, non_strict_class=False):
+    grasp_cls = ds.Designator(cls, name='grasp_cls')
+    support_entity = ds.EdEntityDesignator(robot, id=support, name='support_entity')
+    entity_ids = ds.VariableDesignator([], resolve_type=[ClassificationResult], name='entity_ids')
+    waypoint = ds.EdEntityDesignator(robot, id=waypoint, name='waypoint')
+
+    arm = ds.LockingDesignator(ds.UnoccupiedArmDesignator(robot, {}))
```
LarsJanssenTUe: Please add the requirements on the arm, j.i.c. this is reintroduced that it will not immediately break your code.

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,184 @@
...
+                assert nav_to_start.execute() == 'arrived', "I could not navigate back to the start"
+
+                rospy.logwarn("Robot will turn around to drop the {}".format(grasp_cls.resolve()))
+
+                force_drive = ForceDrive(robot, 0, 0, 1, 3.14)  # rotate 180 degs in pi seconds
+                force_drive.execute()
+
+                say_drop = Say(robot, sentence="I'm going to drop the item, please hold it!")
+                say_drop.execute()
+
```
LarsJanssenTUe: maybe give the operator some time in case he forgot? Like 3 secs?

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,184 @@
...
+                drop_it = SetGripper(robot, arm_designator=arm, grab_entity_designator=grasp_entity)
```
LarsJanssenTUe: There is not really a reason to sent the grasp_entity, but it doesn't hurt anybody either. Up to you!

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,184 @@
...
+                drop_it.execute()
+
+                force_drive_back = ForceDrive(robot, 0, 0, -1, 3.14)  # rotate -180 degs in pi seconds
+                force_drive_back.execute()
```
LarsJanssenTUe: Can't these lines be removed?

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,184 @@
...
+
+    parser = argparse.ArgumentParser(description='Benchmark grasping, for a single item or multiple items at various locations')
+    parser.add_argument("--robot", default="hero",
+                        help="Robot name (amigo, hero, sergio)")
+    parser.add_argument("--output", default="grasp_benchmark.csv",
+                        help="Output of the benchmark results")
+    parser.add_argument("--non-strict-class", action='store_true',
+                        help="Only grasp an item if it matches the specified class. "
+                             "Pass --non-strict-class=False to also grab another class of items "
+                             "if none of the expected class is observed. "
```
LarsJanssenTUe: if no object of the expected class is observed

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,184 @@
...
+    return record
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description='Benchmark grasping, for a single item or multiple items at various locations')
+    parser.add_argument("--robot", default="hero",
+                        help="Robot name (amigo, hero, sergio)")
+    parser.add_argument("--output", default="grasp_benchmark.csv",
+                        help="Output of the benchmark results")
+    parser.add_argument("--non-strict-class", action='store_true',
```
LarsJanssenTUe: If you do not care about object recognition you could also just give a '*' to the cls variable right? Is this not double functionality?

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,184 @@
...
+            results_writer.writeheader()
+
+        robot = get_robot(args.robot)
+
+        if args.subcommand == 'single':
+            single_item(robot, results_writer,
+                        args.cls, args.support, args.waypoint, args.inspect_from_area,
+                        args.non_strict_class)
+        elif args.subcommand == 'batch':
+            with open(args.configuration) as config_file:
```
LarsJanssenTUe: TBH I can only partly follow this read/write shizzle, but I think a config_file is needed, which is not available on git, so that needs to be added

robot_skills/src/robot_skills/mockbot.py
```diff
@@ -233,6 +233,8 @@ def __init__(self, robot_name, tf_listener, *args, **kwargs):
         self._dynamic_entities['john'] = self.generate_random_entity(id='john', type='person')
         self._static_entities = defaultdict(self.generate_random_entity,
                                      {e.id: e for e in [self.generate_random_entity() for _ in range(5)]})
+        self._static_entities['test_waypoint_1'] = self.generate_random_entity(id='test_waypoint_1', type='waypoint')
+        self._static_entities['cabinet'] = self.generate_random_entity(id='cabinet')
```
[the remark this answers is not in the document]
LoyVanBeek: That is only for the ID 'cabinet'. The test in this PR needs a waypoint and those didn't exist here. But there has to be a better way to do all this eventually.

robot_smach_states/.gitignore
```diff
@@ -9,3 +9,4 @@ _readthedocs_build
 autoapi
 _build_rtd
 /.tox/
+benchmark/*.csv
```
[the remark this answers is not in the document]
LoyVanBeek: Those change every time you run the test. Not desirable, but keeping them *somewhere* might be useful. Not sure where though.

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,184 @@
...
+from robot_skills.get_robot import get_robot
+
+# Robot Smach States
+import robot_smach_states.util.designators as ds
+from robot_skills.util.entity import Entity
+from robot_skills.util.kdl_conversions import VectorStamped
+from robot_smach_states import Grab, Inspect, ClassificationResult, NavigateToWaypoint, ForceDrive, SetGripper, \
+    Say
+
+try:
```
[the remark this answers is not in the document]
LoyVanBeek: Indeed. If there is typing, then you get some type hints, otherwise not, but nothing will break.

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,184 @@
...
+        record['inspect_end'] = time.time()
+        record['inspect_duration'] = '{:3.3f}'.format(record['inspect_end'] - record['inspect_start'])
+
+        inspection_result = entity_ids.resolve()  # type: List[ClassificationResult]
+        if inspection_result:
+            if grasp_cls.resolve() == '*' or non_strict_class:
+                rospy.loginfo("Any item will match")
+                matching_results = inspection_result
+            else:
+                matching_results = [result for result in inspection_result if result.type == grasp_cls.resolve()]
```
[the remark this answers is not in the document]
LoyVanBeek: If you want to grasp anything, there is a special arg you can pass. See the argparse bits for this

robot_smach_states/benchmark/benchmark_grasping.py
LoyVanBeek: The entire ID is a long string and useless after the fact. The only reason to keep the ID is to maybe see that the robot tries to take the same object in consecutive tests, but it's useless otherwise. 

So, only the first 6 characters of a single ID are taken

robot_smach_states/benchmark/benchmark_grasping.py
LoyVanBeek: Could be, but descriptive enough IMO. Maybe I make a str with the same length as the float so formatting is nicer

robot_smach_states/benchmark/benchmark_grasping.py
LoyVanBeek: There should be a `HandoverToHuman` in there ideally

robot_smach_states/benchmark/benchmark_grasping.py
LoyVanBeek: This clear the entity from the things the arm is holding.

robot_smach_states/benchmark/benchmark_grasping.py
LoyVanBeek: Nope, this rotates the robot back to it's original orientation

robot_smach_states/benchmark/benchmark_grasping.py
LoyVanBeek: No, there are several cases I should document more clearly.

- Grab a specific class and fail if that is not available
- Grab a specific class and continue with whatever you find if that is not available.
- Grab any class, of course fail if there is nothing available

robot_smach_states/benchmark/benchmark_grasping.py
LoyVanBeek: Yes, hope to make that tomorrow

robot_smach_states/benchmark/benchmark_grasping.py
LoyVanBeek: Yes, some configuration file for a batch of these tests is to be added.

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,184 @@
...
+def single_item(robot, results_writer, cls, support, waypoint, inspect_from_area=None, non_strict_class=False):
+    grasp_cls = ds.Designator(cls, name='grasp_cls')
+    support_entity = ds.EdEntityDesignator(robot, id=support, name='support_entity')
+    entity_ids = ds.VariableDesignator([], resolve_type=[ClassificationResult], name='entity_ids')
+    waypoint = ds.EdEntityDesignator(robot, id=waypoint, name='waypoint')
+
+    arm = ds.LockingDesignator(ds.UnoccupiedArmDesignator(robot, {}))
+    arm.lock()
+
+    record = {'robot': robot.robot_name, 'start_waypoint': waypoint, 'expected_class': cls,
```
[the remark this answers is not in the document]
LoyVanBeek: This is a way to explicitly set this field to be empty if the object was not found

robot_smach_states/benchmark/benchmark_grasping.py
LoyVanBeek: It could with `robot.ed.get_entity(id=...)` but then I'd be replicating what happens in this designator I need anyways

robot_smach_states/.gitignore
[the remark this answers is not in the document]
LoyVanBeek: Maybe MEGA is the best place for that

robot_smach_states/benchmark/benchmark_grasping.py
LoyVanBeek: Any suggestions for what that should all be?

robot_smach_states/benchmark/benchmark_grasping.py
LoyVanBeek: The robot is slow enough actually I think.

robot_smach_states/.gitignore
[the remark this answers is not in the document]
LarsJanssenTUe: I would just remove the gitignore file.

robot_smach_states/benchmark/benchmark_grasping.py
LarsJanssenTUe:
```suggestion
    arm = ds.LockingDesignator(ds.UnoccupiedArmDesignator(robot, arm_properties={"required_trajectories": ["prepare_grasp"], "required_goals": ["carrying_pose"],
                                                               "required_gripper_types":arms.GripperTypes.GRASPING}))
```

Note that arms should also be imported (from robot_skills import arms)

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,224 @@
...
+                        args.cls, args.support, args.waypoint, args.inspect_from_area,
+                        args.non_strict_class)
+        elif args.subcommand == 'batch':
+            with open(args.configuration) as config_file:
+                config_reader = csv.DictReader(config_file)
+                assert config_reader.fieldnames == batch_config_fields, "CSV file need fields {}".format(','.join(batch_config_fields))
+
+                configs = []
+
+                environment_config_description = ""
```
LarsJanssenTUe: What is the use of this? It is only used in line 212 but as far as I see it there is no use...

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,224 @@
...
+                    configs += [config_row]
+                    item_config_description = "Put a {cls} {search_area} the {support}"\
+                        .format(cls=config_row['cls'],
+                                support=config_row['support'],
+                                search_area=config_row['search_area'])
+                    rospy.logwarn(item_config_description)
+                    environment_config_description += item_config_description
+
+                records = []
+                for config_row in configs:
```
LarsJanssenTUe: Isn't this exactly the same as line 205 ("for config_row in config_reader")

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,224 @@
...
+        say_fail.execute()
+    finally:
+        results_writer.writerow(record)
+
+    return record
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description='Benchmark grasping, for a single item or multiple items at various locations'
+                                                 'By specifying ANY as the class, anything class of object will '
```
LarsJanssenTUe: This sentence is a little weird

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,224 @@
...
+                configs = []
+
+                environment_config_description = ""
+                for config_row in config_reader:
+                    configs += [config_row]
+                    item_config_description = "Put a {cls} {search_area} the {support}"\
+                        .format(cls=config_row['cls'],
+                                support=config_row['support'],
+                                search_area=config_row['search_area'])
+                    rospy.logwarn(item_config_description)
```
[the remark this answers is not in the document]
LoyVanBeek: Could be, but this stands out more. It warns the user that something must be done by the user

robot_smach_states/benchmark/benchmark_grasping.py
LoyVanBeek: Maybe the overall description can be removed

robot_smach_states/benchmark/benchmark_grasping.py
LoyVanBeek: Yes. But I want to iterate over the items once first to notify the user and then iterate over them again to do the actual work

robot_smach_states/.gitignore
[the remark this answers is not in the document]
LoyVanBeek: Will put this in git so we can track progress over time as well

LoyVanBeek:
Quite happy with how this works currently. Just used it to see how much difference the PR's in https://github.com/tue-robotics/head_ref/pull/13 make: from mean 10s for inspection and 26s grasping down to 7 and 21 respectively. 
Nice!

robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -0,0 +1,229 @@
...
+
+    return record
+
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description='Benchmark grasping, for a single item or multiple items at various locations.'
+                                                 'By specifying ANY as the class, any class of object will be picked up,'
+                                                 'ignoring the object type'
+                                                 'By specifying a specific class *but* add the --non-strict-class flag, '
```
LarsJanssenTUe: *adding

PR #985, opened 2020-01-14
added default radius as inf

PetervDooren:
merge after https://github.com/tue-robotics/ed/pull/94 and https://github.com/tue-robotics/geolib2/pull/34


robot_skills/src/robot_skills/world_model_ed.py
```diff
@@ -26,6 +26,7 @@
 from robot_skills.util.entity import from_entity_info
 from robot_skills.robot_part import RobotPart
 
+infradius = 200000
```
LarsJanssenTUe: Why this magic number?

PetervDooren: Because while there are symbols for infinity in both python and c++. I don't think it is possible to send it over a rosmessage

MatthijsBurgh: Infiniti can be send in a ros msg.

PetervDooren: Fixed the magic number both here and in https://github.com/tue-robotics/ed/pull/94

PetervDooren:
Added a few more changes to this PR. Changed the emptyspotdesignator to work with our new query strategy. And noticed the example for the place designator did not work so I fixed that

robot_smach_states/src/robot_smach_states/manipulation/place_designator.py
```diff
@@ -89,9 +89,10 @@ def _resolve(self):
         rospy.logerr("Could not find an empty spot")
         return None
 
-    def is_poi_occupied(self, frame_stamped):
+    def is_poi_occupied(self, frame_stamped, surface_entity):
```
LarsJanssenTUe: is_poi_unoccupied

PR #986, opened 2020-01-16
[Do not merge] Fix/gripper interface

LarsJanssenTUe:
Merge after tue-robotics/hero_bridge#12

### DIFF
--- a/robot_skills/src/robot_skills/arms.py
+++ b/robot_skills/src/robot_skills/arms.py
@@ -144,7 +144,7 @@ class PublicArm(object):
 
         return gripper_type is not None and gripper_type in self._available_gripper_types
 
-    def send_gripper_goal(self, state, timeout=5.0, gripper_type=None, max_torque=0.1):
+    def send_gripper_goal(self, state, timeout=5.0, gripper_type=None, max_force=0.1):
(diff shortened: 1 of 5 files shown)


robot_skills/src/robot_skills/arms.py
```diff
@@ -646,7 +646,7 @@ def send_gripper_goal(self, state, timeout=5.0, max_torque=0.1):
             goal.command.direction = GripperCommand.OPEN
         elif state == GripperState.CLOSE:
             goal.command.direction = GripperCommand.CLOSE
-            goal.command.max_torque = max_torque
+            goal.command.max_force = max_force
```
MatthijsBurgh: This is a msg, so change that too.

LarsJanssenTUe: it seems I indeed missed a few things.... not even only the msg.... Sorry for that! Good catch!

LarsJanssenTUe:
Please do not merge yet! I am still in contact with Toyota about this issue!

PetervDooren:
Ping @LarsJanssenTUe, Is this PR something we should still work on sometime? Or shall we close it?

PR #983, opened 2020-01-07
Dont have named arm parts, try2

alberth:
Apparently remaining parts of #717 . I was under the impression this was already merged, and it mostly is indeed, as you can see between the previous and this merge request.

PR #987, opened 2020-01-18
fixed imports of Emptyspotdesignator

PetervDooren:
When merging our emptyspotdesignator we forgot to update some of the imports in challenges. Since the emptyspotdesignator is no longer with the other designators it cannot be imported as ds.Emptyspotdesignator

### DIFF
--- a/challenge_help_me_carry/src/challenge_help_me_carry.py
+++ b/challenge_help_me_carry/src/challenge_help_me_carry.py
@@ -5,6 +5,7 @@ import smach
 import datetime
 import robot_smach_states as states
 import robot_smach_states.util.designators as ds
+from robot_smach_states.manipulation.place_designator import EmptySpotDesignator
 
 from robot_skills.arms import GripperState
 from robocup_knowledge import load_knowledge
@@ -27,11 +28,6 @@ class ChallengeHelpMeCarry(smach.StateMachine):
 
         self.car_waypoint = ds.EntityByIdDesignator(robot, id=challenge_knowledge.waypoint_car['id'])
 
-        self.place_position = ds.LockingDesignator(ds.EmptySpotDesignator(robot, self.target_destination,
-                                                                          name="placement",
-                                                                          area=challenge_knowledge.default_area),
-                                                   name="place_position")
-
         self.empty_arm_designator = ds.UnoccupiedArmDesignator(robot, {}, name="empty_arm_designator")
 
         # With the empty_arm_designator locked, it will ALWAYS resolve to the same arm, unless it is unlocked.
@@ -39,6 +35,14 @@ class ChallengeHelpMeCarry(smach.StateMachine):
 
         self.bag_arm_designator = self.empty_arm_designator.lockable()
         self.bag_arm_designator.lock()
+
+        self.place_position = ds.LockingDesignator(EmptySpotDesignator(robot, self.target_destination,
+                                                                          arm_designator=self.bag_arm_designator,
+                                                                          name="placement",
+                                                                          area=challenge_knowledge.default_area),
+                                                   name="place_position")
+
+
 
         # We don't actually grab something, so there is no need for an actual thing to grab
 
--- a/challenge_help_me_carry/src/challenge_help_me_carry2.py
+++ b/challenge_help_me_carry/src/challenge_help_me_carry2.py
@@ -5,6 +5,7 @@ import smach
 import datetime
 import robot_smach_states as states
 import robot_smach_states.util.designators as ds
+from robot_smach_states.manipulation.place_designator import EmptySpotDesignator
 
 from robot_skills.arms import GripperState
 from robocup_knowledge import load_knowledge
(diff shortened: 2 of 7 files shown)


challenge_help_me_carry/src/challenge_help_me_carry.py
```diff
@@ -27,11 +28,6 @@ def __init__(self, robot):
...
         self.car_waypoint = ds.EntityByIdDesignator(robot, id=challenge_knowledge.waypoint_car['id'])
 
-        self.place_position = ds.LockingDesignator(ds.EmptySpotDesignator(robot, self.target_destination,
-                                                                          name="placement",
-                                                                          area=challenge_knowledge.default_area),
-                                                   name="place_position")
-
         self.empty_arm_designator = ds.UnoccupiedArmDesignator(robot, {}, name="empty_arm_designator")
 
         # With the empty_arm_designator locked, it will ALWAYS resolve to the same arm, unless it is unlocked.
```
LarsJanssenTUe: Please also move these lines

challenge_help_me_carry/src/challenge_help_me_carry2.py
```diff
@@ -27,11 +28,6 @@ def __init__(self, robot):
...
         self.car_waypoint = ds.EntityByIdDesignator(robot, id=challenge_knowledge.waypoint_car['id'])
 
-        self.place_position = ds.LockingDesignator(ds.EmptySpotDesignator(robot, self.target_destination,
-                                                                          name="placement",
-                                                                          area=challenge_knowledge.default_area),
-                                                   name="place_position")
-
         self.empty_arm_designator = ds.UnoccupiedArmDesignator(robot, {}, name="empty_arm_designator")
 
         # With the empty_arm_designator locked, it will ALWAYS resolve to the same arm, unless it is unlocked.
```
LarsJanssenTUe: Same holds for these lines

challenge_storing_groceries/src/challenge_storing_groceries/manipulate_machine.py
```diff
@@ -82,7 +83,7 @@ def __init__(self, robot, grab_designator=None):
         smach.StateMachine.__init__(self, outcomes=["succeeded", "failed"])
 
         # Create designators
-        self.empty_arm_designator = ds.UnoccupiedArmDesignator(robot, {}, name="empty_arm_designator")
+        self.empty_arm_designator = ds.OccupiedArmDesignator(robot, {}, name="empty_arm_designator")
```
LarsJanssenTUe: Why change this from Unoccupied to Occupied?

challenge_set_a_table/src/challenge_set_a_table_states/manipulate_machine.py
```diff
@@ -178,9 +179,11 @@ def __init__(self, robot, grasp_designator1, grasp_designator2, grasp_designator
         grasp_furniture_designator2 = ds.EntityByIdDesignator(robot, id=grasp_furniture_id2)
 
         place_furniture_designator = ds.EntityByIdDesignator(robot, id=place_furniture_id)
-        place_designator = ds.EmptySpotDesignator(robot=robot,
-                                                  place_location_designator=place_furniture_designator,
-                                                  area="on_top_of")
+        arm_designator = ds.OccupiedArmDesignator(robot, {})
```
LarsJanssenTUe: Why an OccupiedArmDesignator? All the others have an UnOccupiedArmDesignator

challenge_help_me_carry/src/challenge_help_me_carry.py
PetervDooren: Why? these lines are about the empty arm designator. The new API of the emptyspotdesignator requires an arm so I just moved it to after the definition of the arm_designator

challenge_storing_groceries/src/challenge_storing_groceries/manipulate_machine.py
PetervDooren: Because I did an oopsie

challenge_set_a_table/src/challenge_set_a_table_states/manipulate_machine.py
PetervDooren: Depends on the situation. In this case however there is no reason to choose either an occupied or unoccupied designator. So I changed it to a general armdesignator. This could be integrated better in the challenge if we know which arm we will use for the operation but that would require API breaks in the challenge. I think we should only look at that if we start working on this challenge.

challenge_help_me_carry/src/challenge_help_me_carry.py
LarsJanssenTUe: That sounds good!

challenge_set_a_table/src/challenge_set_a_table_states/manipulate_machine.py
LarsJanssenTUe: Fair enough IMO.

LarsJanssenTUe:
CI seems to fail after merging #983. I will look at that on tuesday, after that I think we can merge

PetervDooren:
Agreed

PR #996, opened 2020-02-01
Add robot.reset_all_arms function.

alberth:
Add robot function to reset all arms

### DIFF
--- a/robot_skills/src/robot_skills/robot.py
+++ b/robot_skills/src/robot_skills/robot.py
@@ -133,16 +133,30 @@ class Robot(object):
                 bodypart.reset()
         return all(results.values())
 
+    def reset_all_arms(self, arm_timeout=0.0, close_gripper=True, gripper_timeout=None):
+        """
+        Reset all arms of the robot, and by default, close their grippers.
+
+        The arm is always reset, the gripper is closed if 'close_gripper' is True (which is the default).
+
+        :param arm_timeout: How long to wait for resetting the arm, by default 0.0
+        :param close_gripper: By default 'True', which closes the gripper. If False, the gripper is not activated,
+        :param gripper_timeout: If specified and 'close_gripper' holds, timeout for closing the gripper.
+            If not specified, gripper movement uses 'arm_timeout'.
+        """
+        for arm in self._arms.itervalues():
+            arm.reset(timout=arm_timeout)
+            if close_gripper:
+                if gripper_timeout is None:
+                    gripper_timeout = arm_timeout
+                arm.send_gripper_goal('close', timeout=gripper_timeout)
+
     def standby(self):
         if not self.robot_name == 'amigo':
             rospy.logerr('Standby only works for amigo')
             return
 
-        for arm in self._arms.itervalues():
-            arm.reset()
-        for arm in self._arms.itervalues():
-            arm.send_gripper_goal('close')
-
+        self.reset_all_arms()
         self.head.look_down()
         self.torso.low()
         self.lights.set_color(0, 0, 0)
--- a/robot_skills/src/robot_skills/robot.py
+++ b/robot_skills/src/robot_skills/robot.py
@@ -145,11 +145,12 @@ class Robot(object):
             If not specified, gripper movement uses 'arm_timeout'.
         """
         for arm in self._arms.itervalues():
-            arm.reset(timout=arm_timeout)
             if close_gripper:
                 if gripper_timeout is None:
                     gripper_timeout = arm_timeout
                 arm.send_gripper_goal('close', timeout=gripper_timeout)
+
+            arm.reset(timout=arm_timeout)
 
     def standby(self):
         if not self.robot_name == 'amigo':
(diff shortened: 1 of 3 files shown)


robot_skills/src/robot_skills/robot.py
```diff
@@ -131,16 +131,30 @@ def reset(self):
...
+        The arm is always reset, the gripper is closed if 'close_gripper' is True (which is the default).
+
+        :param arm_timeout: How long to wait for resetting the arm, by default 0.0
+        :param close_gripper: By default 'True', which closes the gripper. If False, the gripper is not activated,
+        :param gripper_timeout: If specified and 'close_gripper' holds, timeout for closing the gripper.
+            If not specified, gripper movement uses 'arm_timeout'.
+        """
+        for arm in self._arms.itervalues():
+            arm.reset(timout=arm_timeout)
+            if close_gripper:
```
LarsJanssenTUe: Why first reset before closing the gripper? In terms of safety I think it might be better to first close the gripper (if necessary) and then reset.

robot_smach_states/src/robot_smach_states/reset.py
```diff
@@ -75,9 +72,7 @@ def __init__(self, robot, timeout=0.0):
         smach.State.__init__(self, outcomes=["done"])
 
     def execute(self, userdata=None):
-        for arm in self.robot.arms.itervalues():
-            arm.reset()
-            arm.send_gripper_goal('close', timeout=self.timeout)
+        self.robot.reset_all_arms(timeout=self.timeout)
```
LarsJanssenTUe: This changes the code, I am not sure if this would change anything, but just to be sure I would prefer the following:

```suggestion
        self.robot.reset_all_arms(gripper_timout=self.timeout)
```

robot_smach_states/src/robot_smach_states/reset.py
```diff
@@ -91,9 +86,7 @@ def __init__(self,robot, timeout=0.0):
         self.timeout = timeout
 
     def execute(self, userdata=None):
-        for arm in self.robot.arms.itervalues():
-            arm.reset()
-            arm.send_gripper_goal('close', timeout=self.timeout)
+        self.robot.reset_all_arms(timeout=self.timeout)
```
LarsJanssenTUe: Same as above

robot_skills/src/robot_skills/robot.py
```diff
@@ -131,16 +131,30 @@ def reset(self):
                 bodypart.reset()
         return all(results.values())
 
+    def reset_all_arms(self, arm_timeout=0.0, close_gripper=True, gripper_timeout=None):
```
LarsJanssenTUe: If I am correct the arm_time follows the following path:
arm.reset --> arms.reset --> send_joint_goal --> send_joint_trajectory --> goal_time_tolerance.

The goal_time_tolerance should not be 0.0 seconds by default right?

alberth: Looks like it, but it there is no explanation of what the timeout is about. I grabbed `0.0` from the `Arm.reset` function, lower level functions seem to have other defaults for unknown reason.

In all checked out repositories at my system (but that is incomplete), the only other mention of `goal_time_tolerance` is in `control_msgs/control_msgs/action/FollowJointTrajectory.action`:
```
# To report success, the joints must be within goal_tolerance of the
# final trajectory value.  The goal must be achieved by time the
# trajectory ends plus goal_time_tolerance.  (goal_time_tolerance
# allows some leeway in time, so that the trajectory goal can still
# succeed even if the joints reach the goal some time after the
# precise end time of the trajectory).
#
# If the joints are not within goal_tolerance after "trajectory finish
# time" + goal_time_tolerance, the goal aborts with error_code set to
# GOAL_TOLERANCE_VIOLATED
```
That reads as additional time on top of the normally required time. For a reliable system, `0.0` should be fine in that case.

robot_skills/src/robot_skills/robot.py
alberth: good point, changed

robot_smach_states/src/robot_smach_states/reset.py
alberth: fixed

robot_smach_states/src/robot_smach_states/reset.py
alberth: fixed

alberth:
@LarsJanssenTUe  The only remaining point is the timeout. Currently, `reset_all_arms` does that `arm.reset` does too.

robot_skills/src/robot_skills/robot.py
LarsJanssenTUe: This description indeed seems to indicate that a default of 0.0 is fine!

robot_skills/src/robot_skills/robot.py
```diff
@@ -133,16 +133,31 @@ def reset(self):
...
+        :param gripper_timeout: If specified and 'close_gripper' holds, timeout for closing the gripper.
+            If not specified, gripper movement uses 'arm_timeout'.
+        """
+        for arm in self._arms.itervalues():
+            if close_gripper:
+                if gripper_timeout is None:
+                    gripper_timeout = arm_timeout
+                arm.send_gripper_goal('close', timeout=gripper_timeout)
+
+            arm.reset(timout=arm_timeout)
```
LarsJanssenTUe:
```suggestion
            arm.reset(timeout=arm_timeout)
```

LarsJanssenTUe:
Tested and after fixing a small typo it seems to work

PR #1004, opened 2020-02-18
Receptionist 2020: sync code & check for people in seats, not just volume

LoyVanBeek:


robot_smach_states/src/robot_smach_states/manipulation/point_at.py
```diff
@@ -20,8 +20,9 @@ def __init__(self, robot, arm_designator, point_at_designator, look_at_designato
         Drive the robot back a little and move the designated arm to place the designated item at the designated pose
 
         :param robot: Robot to execute state with
-        :param arm_designator : arm to place with, so Arm that holds entity_to_place, E.g. ArmHoldingEntityDesignator
+        :param arm_designator : arm to point with, so eg. an UnoccupiedArmDesignator
```
PetervDooren:
```suggestion
        :param arm_designator : arm to point with, so eg. an ArmDesignator
```

robot_smach_states/src/robot_smach_states/world_model.py
```diff
@@ -223,7 +226,7 @@ class CheckEmpty(smach.State):
     """
     Check whether a volume of an entity is filled
     """
-    def __init__(self, robot, segmented_entity_ids_designator, entity_designator, volume, threshold=None):
+    def __init__(self, robot, segmented_entity_ids_designator, entity_designator, volume_designator, threshold=None):
```
PetervDooren: Since volume can also be a regular string I would keep the variable name as is. Just update the docstring

robot_smach_states/src/robot_smach_states/world_model.py
```diff
@@ -237,20 +240,37 @@ def __init__(self, robot, segmented_entity_ids_designator, entity_designator, vo
...
         self.entity_des = entity_designator
-        self.volume = volume
+
+        ds.check_type(volume_designator, str)
+        if isinstance(volume_designator, str):
+            self.volume_designator = ds.VariableDesignator(volume_designator)
+        elif isinstance(volume_designator, ds.Designator):
+            self.volume_designator = volume_designator
+        else:
+            raise RuntimeError("This shouldn't happen. Wrong types should have raised an exception earlier")
```
PetervDooren: I think this is a bit backwards. If I input a string now it will be transformed to a designator only to be turned back into a string when I need it.

Also, I believe that we do this 'it can be a thing or a designator to a thing' often enough that it warrants its own support function much like check_type.

robot_smach_states/src/robot_smach_states/world_model.py
```diff
@@ -237,20 +240,37 @@ def __init__(self, robot, segmented_entity_ids_designator, entity_designator, vo
...
-                rospy.loginfo('Occupied space is {}, remaining space is {}'.format(occupied_space, remaining_space))
-                if remaining_space > self.threshold:
+                ratio = occupied_space/vol.size
+                rospy.loginfo('{n}: vol.size: {s:.3f}m^3, Occupied space is {o:.3f}m^3, remaining space is {r:.3f}m^3. Occupied/vol.size={rat:.3f}% filled, threshold to mark as occupied is {t:.3f}%'.format(n=volume_name,
+                    s=vol.size,
+                    o=occupied_space,
+                    r=remaining_space,
+                    t=self.threshold*100,
+                    rat=ratio*100))
+                if ratio > self.threshold:
```
PetervDooren: Update the docstring for threshold if you change the use of it

challenge_receptionist/src/challenge_receptionist/find_empty_seat.py
```diff
@@ -37,6 +50,75 @@ def __repr__(self):
...
+        if self.debug:
+            import ipdb;ipdb.set_trace()
+
+        try:
+            seat_entity = self.seat_entity_designator.resolve()  # type: Entity
+            seat_volume_names = sorted([k for k in seat_entity.volumes.keys() if k.endswith('seat')])
+
+            return seat_volume_names
+        except AttributeError:
+            return None
```
PetervDooren: Can be made shorter

if seat_entity:
    seat_volume_na.....

challenge_receptionist/src/challenge_receptionist/find_empty_seat.py
```diff
@@ -46,77 +128,79 @@ class FindEmptySeat(smach.StateMachine):
...
-                                   states.SayFormatted(robot,
-                                                       ["Let me find a place for {name} to sit. Please be patient while I check out where there's place to sit"],
-                                                       name=seat_is_for,
-                                                       block=False),
-                                   transitions={'spoken': 'ITERATE_NEXT_SEAT'})
-
-            smach.StateMachine.add('ITERATE_NEXT_SEAT',
-                                   states.IterateDesignator(seats, seat_ent_des.writeable),
-                                   transitions={'next': 'CHECK_SEAT_EMPTY',
-                                                'stop_iteration': 'SAY_NO_EMPTY_SEATS'})
```
PetervDooren: By removing the iteration over the different seats you assume that you will find all people in the room in one go. I am not saying that it's a bad assumption, just be aware that you made it

challenge_receptionist/src/challenge_receptionist/find_empty_seat.py
```diff
@@ -19,8 +29,11 @@ def __init__(self, robot, seat_ids, room, name=None):
 
         self.room = room
         self.seat_ids = seat_ids
+        self.debug = debug
 
     def _resolve(self):
+        if self.debug:
+            import ipdb;ipdb.set_trace()
```
PetervDooren: Do you want to keep the debug stuff here? or did you forget to delete it?

challenge_receptionist/src/challenge_receptionist/find_empty_seat.py
```diff
@@ -1,15 +1,25 @@
...
 import robot_smach_states.util.designators as ds
 import smach
 from robot_skills.util.entity import Entity
 from robot_skills.util.volume import Volume
 from robot_skills.classification_result import ClassificationResult
 
+try:
+    from typing import List
+except ImportError:
+    pass
```
PetervDooren: What is this? and why does it get a pass if it does not import?

robot_smach_states/src/robot_smach_states/world_model.py
LoyVanBeek: Agreed.

robot_smach_states/src/robot_smach_states/util/designators/utility.py
```diff
@@ -8,10 +8,28 @@
 import operator
 
 # TU/e Robotics
-from .core import Designator, VariableDesignator
+from robot_smach_states.util.designators.core import Designator, VariableDesignator
 
 __author__ = 'loy'
 
+def resolve(designator_or_value):
```
PetervDooren: Beautiful! After this PR we should look where we should use this in robot_smach_states.

MatthijsBurgh: I don't agree with this. This is only needed in case of bad management of your variables. This allows for creating a bigger mess, while we should reduce the number of possible variable types.

robot_smach_states/src/robot_smach_states/util/designators/utility.py
```diff
@@ -8,10 +8,28 @@
 import operator
 
 # TU/e Robotics
-from .core import Designator, VariableDesignator
+from robot_smach_states.util.designators.core import Designator, VariableDesignator
```
MatthijsBurgh: Why??

robot_smach_states/src/robot_smach_states/util/designators/utility.py
PetervDooren: I disagree. This designator or value stuff already happens a lot. Mostly for things that might be but don't have to be defined at construction time. For example with area names

MatthijsBurgh: Jep, but that doesn't mean we should continue in the wrong direction. In case that the value is already known on construction you could put the value in a designator, therefore you always have a designator and never a value.

PetervDooren: I still disagree. But I think this is a discussion to have on a tuesday evening.

robot_smach_states/src/robot_smach_states/util/designators/utility.py
LoyVanBeek: My Python install complained about this not happening in a package.

robot_smach_states/src/robot_smach_states/util/designators/utility.py
LoyVanBeek: You both have a point here. 
This `resolve` function allows to more easily upgrade a parameter to be a Designator later on, but otherwise I agree it should not be necessary.

robot_smach_states/src/robot_smach_states/util/designators/utility.py
MatthijsBurgh: Weird

robot_smach_states/src/robot_smach_states/util/designators/utility.py
LoyVanBeek: The difficult thing here is that oftentimes, you only find out something should have been runtime variable at a later time. To not change the external interface, you modify the internals to accept both a concrete value *and* Designator. This is a pattern I've seen a few times at least. 
So, reusing the code for that pattern is handy. 

The other option is to make larger changes to a constructor and all it's uses each time we find that a parameter should actually be runtime variable. But this is error-prone...

challenge_receptionist/src/challenge_receptionist/find_empty_seat.py
LoyVanBeek: These are used only for type hinting (in PyCharm etc.). No runtime code breaks if this is missing.

challenge_receptionist/src/challenge_receptionist/find_empty_seat.py
LoyVanBeek: I wantto keep it, it's optional and off by default

challenge_receptionist/src/challenge_receptionist/find_empty_seat.py
LoyVanBeek: Better safe than sorry

challenge_receptionist/src/challenge_receptionist/find_empty_seat.py
LoyVanBeek: True. The state for finding ppl in room takes care of looking around.

robot_smach_states/src/robot_smach_states/util/designators/utility.py
LoyVanBeek: I've updated this PR to match my advice at the bottom of my comment. 

The ease is nice, but clarity is more important.

MatthijsBurgh: I still don't like the function. This way you are facilitating the bad habits. I don't say we should change the old code, but we are now aware of the pattern. So we should be forced to think more about the design, before we start coding. Adding this function doesn't force us in that direction.

LoyVanBeek: I'll remove the function as well, gimme a sec.

robot_smach_states/src/robot_smach_states/world_model.py
```diff
@@ -233,8 +232,9 @@ def __init__(self, robot, segmented_entity_ids_designator, entity_designator, vo
 
         :param segmented_entity_ids_designator: designator containing the segmented objects in the volume
         :param entity_designator: EdEntityDesignator indicating the (furniture) object to check
-        :param volume_designator: string (or Designator resolving to a str) defining which volume of the entity is checked
-        :param threshold: float [m^3] indicating the free volume above which the area is considered partially_occupied.
+        :param volume_designator: string-Designator defining which volume of the entity is checked
+        :param threshold: float indicating the ration between free and occupied space in the volume.
```
PetervDooren:
```suggestion
        :param threshold: float indicating the ratio between free and occupied space in the volume.
```

robot_smach_states/src/robot_smach_states/world_model.py
```diff
@@ -243,15 +243,19 @@ def __init__(self, robot, segmented_entity_ids_designator, entity_designator, vo
...
 
         ds.check_type(volume_designator, str)
-        # Can be either designator or concrete value, but the resolve function takes care of either case
-        self.volume_designator = volume_designator
+        if isinstance(volume_designator, str):
+            self.volume_designator = ds.VariableDesignator(volume_designator)
+        elif isinstance(volume_designator, ds.Designator):
+            self.volume_designator = volume_designator
+        else:
+            raise RuntimeError("This shouldn't happen. Wrong types should have raised an exception earlier")
```
PetervDooren: If we want to enforce that only designators are used for this function we should not add support for string inputs. In that case even the check_type above this should be replaced with
assert(volume_designator.resolve_type is <str>)

robot_smach_states/src/robot_smach_states/util/designators/utility.py
PetervDooren: 1800 seconds have passed Loy. Please remove this function before merging

PetervDooren:
@LoyVanBeek what do you want to do with this PR?

LoyVanBeek:
I'd close it. We haven't missed it in ~1.5 years, so once we revisit this challenge we can re-open (if we remember to) and process the last comments or re-implement with newer insights and updated challenge.

MatthijsBurgh:
We more of these PRs we should collect them @PetervDooren

PR #1006, opened 2020-03-01
Navigate using designators

PetervDooren:
We Use a lot of different NavigateTo smach states. While this is understandable they often contain similar navigation constraints. Such as in place and grasp and almost every state looks at an entity.
Furthermore this will allow us to mix constraints where relevant. So we are able to say "You need to be in the living room and able to see the dinnertable"


robot_smach_states/src/robot_smach_states/navigation/navigate_to_designator.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class NavigateToDesignator(NavigateTo):
+    def __init__(self, robot, constraint_designator, reset_head=True, speak=True):
+        """
+        @param constraint_designator a Designator that resolves to navigation constraints
+        """
+        super(NavigateToDesignator, self).__init__(robot, reset_head=reset_head, speak=speak)
+
+        check_resolve_type(constraint_designator, tuple)  #TODO verify in more detail that this is indeed a constraint designator
```
PetervDooren: @LoyVanBeek As designator expert, do you think there is a way to verify that the constraint resolves to a tuple containing a Positionconstraint and an Orientationconstraints?

LoyVanBeek: I don't think so, as tuples are interpreted as something else.
But the neater way to do this might be to return a custom type, eg. a `PoseConstraint` which is composed of both a `PositionConstraint` and a `OrientationConstraint`. 
Will also clean up some type hinting.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -278,6 +280,30 @@ def distance(e1, e2):
...
+    """
+    def __init__(self, robot, entity_designator, arm_designator=None):
+        constraint_designator = CompoundConstraintsDesignator()
+        constraint_designator.add(ArmsreachConstraintsDesignator(robot, entity_designator, arm_designator, look=False), 'reach')
+        constraint_designator.add(LookAtConstraintsDesignator(entity_designator, 1.57), 'face') #TODO hero-specific magic number
+        super(NavigateToPreGrasp, self).__init__(robot, constraint_designator, reset_head=True)
+
+
+class NavigateToGrasp(NavigateToDesignator):
+    """Navigate so that the arm can easily grasp the entity
```
jlunenburg: Basically, this class replaces some other class. I think we should remove the other one, otherwise it will be unclear which one to use in the future

robot_smach_states/src/robot_smach_states/manipulation/place_designator.py
```diff
@@ -176,14 +176,20 @@ def _create_selection_marker(self, selected_pose):
         marker.lifetime = rospy.Duration(30.0)
         return marker
 
-    def _determine_points_of_interest_with_area(self, entity, area):
+    def _determine_points_of_interest_with_area(self, entity, areades):
         """ Determines the points of interest using an area
         :type entity: Entity
-        :param area: str indicating which volume of the entity to look at
+        :param area: str or str Designator indicating which volume of the entity to look at
```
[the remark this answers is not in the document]
PetervDooren: Sometimes we may not know which area to place on at the time of construction. In that case we have no choice but to pass a designator to the Emptyspotdesignator

robot_smach_states/src/robot_smach_states/util/designators/ed_designators.py
```diff
@@ -27,7 +27,7 @@ class EdEntityCollectionDesignator(Designator):
     >>> check_resolve_type(entities, [Entity]) #This is more a test for check_resolve_type to be honest :-/
     """
 
-    def __init__(self, robot, type="", center_point=None, radius=0, id="", parse=True, criteriafuncs=None,
+    def __init__(self, robot, type="", center_point=None, radius=float('inf'), id="", parse=True, criteriafuncs=None,
```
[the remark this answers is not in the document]
PetervDooren: We made a change in the ed_simple_query. It used to be that radius=0 signaled that position should not be taken into account. This change keeps the designator in line with world_model.py in robot_skills.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
PetervDooren: The other one (navigate_to_grasp.py) is already gone

robot_smach_states/src/robot_smach_states/manipulation/place_designator.py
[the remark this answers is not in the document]
LoyVanBeek: You can make the `__init__` smart enough to accept both the str and a designator to a str.

LoyVanBeek:
> designators; these are originally intended to store data

Not entirely true: designators are meant to designate a concrete value which is described at programming-time and resolved to a concrete value at runtime. Eg.g describe an entity matching some criteria and resolve to a specific entity. 
Directly writing to them is more of an afterthought than an original purpose.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -278,6 +280,30 @@ def distance(e1, e2):
...
+    :param arm_designator: which arm to eventually grasp with?
+    """
+    def __init__(self, robot, entity_designator, arm_designator=None):
+        constraint_designator = CompoundConstraintsDesignator()
+        constraint_designator.add(ArmsreachConstraintsDesignator(robot, entity_designator, arm_designator, look=False), 'reach')
+        constraint_designator.add(LookAtConstraintsDesignator(entity_designator, 1.57), 'face') #TODO hero-specific magic number
+        super(NavigateToPreGrasp, self).__init__(robot, constraint_designator, reset_head=True)
+
+
+class NavigateToGrasp(NavigateToDesignator):
```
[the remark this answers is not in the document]
PetervDooren: Making it slightly less verbose is worth it in my opinion since several challenges make use of it. And while they may not be very verbose now that might change if we plan to utilise more constraints, e.g. the fact that the robot needs line-of-sight to the object.

You are correct on the 'reset_head' parameter, I will expose that

robot_smach_states/src/robot_smach_states/manipulation/place_designator.py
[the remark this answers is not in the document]
PetervDooren: I added type checks to the __init__ function.

robot_smach_states/src/robot_smach_states/manipulation/place_designator.py
```diff
@@ -176,14 +176,20 @@ def _create_selection_marker(self, selected_pose):
...
         :rtype: [FrameStamped]
         """
 
         # We want to give it a convex hull using the designated area
+        if isinstance(areades, str):
+            area = areades
+        elif areades.resolve_type == str:
+            area = areades.resolve()
+        else:
+            raise AssertionError("Cannot place in {}".format(areades))
```
PetervDooren: Do we have a support fuction for designators which does this? So if areades is already a string just copy it and if it is a designator resolve it? I can image this would be useful in other situations as well

robot_smach_states/src/robot_smach_states/navigation/navigate_to_explore.py
```diff
@@ -80,8 +80,8 @@ def generateConstraint(self):
 
         pc = PositionConstraint(constraint=pci, frame="/map")
         oc = OrientationConstraint(look_at=Point(x, y, 0.0), frame="/map")
-
-        return pc, oc
+        constraint = PoseConstraint(pc, oc)
```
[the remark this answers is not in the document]
PetervDooren: I have now implemented this as a named tuple. Sadly we can't typecheck the two values individually so people can theoretically place anything in the pc and oc fields. But I assume that if they import the tuple they know what it does

robot_smach_states/src/robot_smach_states/manipulation/place_designator.py
LoyVanBeek: For checking, there is `check_type` in the `designator.checks`-submodule or thereabouts. 
To resolve either a designator or just get the already set concrete value, I don't think there is a util yet for that.

robot_smach_states/src/robot_smach_states/util/designators/navigation/waypoint_constraints.py
```diff
@@ -0,0 +1,63 @@
...
+from ..checks import check_resolve_type
+from cb_planner_msgs_srvs.msg import OrientationConstraint, PositionConstraint
+
+
+class WayPointConstaintsDesignator(Designator):
+    def __init__(self, robot, waypoint_designator, radius=0.15, look_at_designator=None, name=None):
+        super(WayPointConstaintsDesignator, self).__init__(resolve_type=tuple, name=name)
+        self.robot = robot
+
+        check_resolve_type(waypoint_designator, Entity)  # Check that the waypoint_designator resolves to an Entity
```
alberth: Repeating in comment what the code says already is useless at best (and worse if code and comment disagree).

Let the code say what is being called or computed at its line, use comment for higher level notions that the code doesn't say.

Examples of higher level notions are for example the aim of the next block of code, or why this step must be done here (and not elsewhere), or why the seemingly weird piece of code is actually correct.

robot_smach_states/src/robot_smach_states/util/designators/navigation/radius_constraints.py
```diff
@@ -0,0 +1,78 @@
...
+            ch = []
+
+        x = e.pose.frame.p.x()
+        y = e.pose.frame.p.y()
+        assert e.pose.frame_id.strip("/") == "map", "ObserveConstraintsDesignator assumes entities are defined w.r.t. map frame"
+
+        if len(ch) > 0:  # If a convex hull is present, use this to create the position constraint
+            pci = ""  # Create an empty position constraint
+
+            for i in xrange(len(ch)):  # Iterate over the convex hull
```
jlunenburg: Use enumerate?

robot_smach_states/src/robot_smach_states/navigation/navigate_to_symbolic.py
```diff
@@ -23,87 +18,27 @@ def __init__(self, robot, entity_designator_area_name_map, entity_lookat_designa
...
-        check_resolve_type(entity_lookat_designator, Entity)  # Check that the entity_designator resolves to an Entity
-        self.entity_lookat_designator = entity_lookat_designator
-
-    def generateConstraint(self):
-        """
-        Generates the position constraint
-        """
-        return self.generate_constraint(self.robot, self.entity_designator_area_name_map, self.entity_lookat_designator)
-
-    @staticmethod
```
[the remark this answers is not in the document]
PetervDooren: I did and I changed the references to refer to the correct function

robot_smach_states/src/robot_smach_states/navigation/navigate_to_waypoint.py
```diff
@@ -1,59 +1,17 @@
...
-            else:
-                rospy.logerr("NavigateToWaypoint::generateConstraint: No entity could be resolved from designator '%s'" % self.look_at_designator)
-
-        if not oc:
-            oc = OrientationConstraint(look_at=Point(x+10, y, 0.0), angle_offset=rz, frame="/map")
-
-        return pc, oc
+        if look_at_designator:
+            constraint_designator = CompoundConstraintsDesignator()
+            constraint_designator.add(WayPointConstraintsDesignator(waypoint_designator, radius, look=False), 'waypoint')
```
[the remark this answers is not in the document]
PetervDooren: No. because navigateToWaypoint will adopt the orientation of the waypoint if no lookat designator is provided.

robot_smach_states/examples/util/designators/navigation/example_arm_constraints_designator.py
```diff
@@ -0,0 +1,38 @@
...
+    robot = get_robot(args.robot)
+
+    # get an armdesignator, entitydesignator and
+    arm = ArmDesignator(robot, {}, name="arm_designator")
+    frame = frame_stamped("\map", 0, 0, 1.3)
+    framedes = VariableDesignator(frame, name="frame designator")
+
+    # create an armsreach constraintdesignator
+    # look at the frame
+    nav1 = ArmsreachConstraintsDesignator(robot, framedes, arm, name="armsreach designator")
```
LoyVanBeek: `ArmsreachConstraintsDesignator` --> `ArmsReachConstraintsDesignator` maybe?

PetervDooren: yeah

PetervDooren:
I added tests and examples for all constraint generations. The only ones I couldnt design tests for were symbolic constraints and radiusconstraints as they both need a worldmodel to be running.

PetervDooren:
After some discussion we have decided to give up on this approach and instead take the approach of #1028

PR #1010, opened 2020-03-07
Grasping object high

PetervDooren:


robot_skills/src/robot_skills/hero.py
```diff
@@ -109,3 +106,40 @@ def move_to_hmi_pose(self):
...
+        z_head = grasp_target.z() + z_over
+
+        if z_head < 1.3:
+            # we dont need to do stupid stuff
+            arm.send_joint_trajectory('prepare_grasp')
+            return True
+
+        # saturate the arm lift goal
+        z_arm = (z_head - z_hh) * torso_to_arm_ratio
+        z_arm = min(0.69, max(z_arm, 0.0))
```
LarsJanssenTUe: Same here: duplicate code

robot_skills/src/robot_skills/hero.py
```diff
@@ -109,3 +106,40 @@ def move_to_hmi_pose(self):
...
+
+        pose = arm._arm.default_trajectories['prepare_grasp']
```
LarsJanssenTUe: This is not the official way of doing it! Use the public arm!

robot_skills/src/robot_skills/hero.py
```diff
@@ -109,3 +106,40 @@ def move_to_hmi_pose(self):
...
+
+        angle_offset = -math.atan2(arm.base_offset.y(), arm.base_offset.x())
+        self.base.turn_towards(grasp_target.x(), grasp_target.y(), "/map", angle_offset)
+        arm.wait_for_motion_done()
+        while self.base.local_planner.getStatus() != "arrived":
+            rospy.sleep(0.1)
+        return True
+
+    def go_to_driving_pose(self):
+        arm = self.parts['leftArm']
```
LarsJanssenTUe: This is a firm hardcode, you code use the get arm with the carrying_pose as a prerequisite

robot_skills/src/robot_skills/hero.py
```diff
@@ -109,3 +106,40 @@ def move_to_hmi_pose(self):
...
+        z_arm = min(0.69, max(z_arm, 0.0))
+
+        pose = arm._arm.default_trajectories['prepare_grasp']
+        pose[1][0] = z_arm
+        arm._arm._send_joint_trajectory(pose)
+
+        angle_offset = -math.atan2(arm.base_offset.y(), arm.base_offset.x())
+        self.base.turn_towards(grasp_target.x(), grasp_target.y(), "/map", angle_offset)
+        arm.wait_for_motion_done()
+        while self.base.local_planner.getStatus() != "arrived":
```
LarsJanssenTUe: Same as before, add not rospy.shutdown. What happens if the robot does not get where it is supposed to go

robot_skills/src/robot_skills/sergio.py
```diff
@@ -39,3 +39,15 @@ def __init__(self, wait_services=False):
         self.add_body_part('ed', world_model_ed.ED(self.robot_name, self.tf_listener))
 
         self.configure()
+
+    def move_to_pregrasp_pose(self, arm, grasp_target):
```
LarsJanssenTUe: Same things as for the Amigo object

robot_smach_states/src/robot_smach_states/navigation/navigation.py
```diff
@@ -103,6 +104,10 @@ def execute(self, userdata=None):
         if self.reset_head:
             self.robot.head.close()
 
+        # Move the robot to a suitable driving pose
+        if self.reset_pose and self.robot.base.global_planner.path_length > 0.5:
```
LarsJanssenTUe: Hmmm this magic number is not really nice here, however I can not really think of a better solution so I guess leave it in for now?

robot_skills/src/robot_skills/base.py
```diff
@@ -198,6 +198,18 @@ def move(self, position_constraint_string, frame):
 
         return plan
 
+    def turn_towards(self, x, y, frame, offset=0.0):
```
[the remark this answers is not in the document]
PetervDooren: What else would you point towards?

robot_skills/src/robot_skills/hero.py
PetervDooren: Not possible, we cannot get the values of the default trajectories from the publicArm. And we shouldn't because normally you would not want to use them for anything.

robot_skills/src/robot_skills/hero.py
```diff
@@ -109,3 +106,40 @@ def move_to_hmi_pose(self):
...
+        This poses the robot for an inspect.
+
+        :param grasp_target: kdl.Frame with the pose of the entity to be grasp.
+        :return: boolean, false if something went wrong.
+        """
+        # calculate the arm_lift_link which must be sent
+        z_over = 0.4  # height the robot should look over the surface
+        z_hh = 0.9  # height of the robots head at z_arm=0
+        torso_to_arm_ratio = 2.0  # motion of arm/motion of torso
+        z_head = grasp_target.z() + z_over
```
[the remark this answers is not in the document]
PetervDooren: I would keep all the crappy code in one function. Let's not pretend that this is generic in any way.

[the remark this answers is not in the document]
LarsJanssenTUe: But now the hardcoded code is used in two different functions, so not only do we have hardcoded information but we have it twice.

robot_skills/src/robot_skills/sergio.py
LarsJanssenTUe: Please change the docstring here as well

PR #1013, opened 2020-03-07
Test/grasp heights

jlunenburg:
Contains test and some 'fixes' in the grasp point determination to enable this.

PR #1016, opened 2020-03-07
Feature/arm requirements

LarsJanssenTUe:


robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -36,6 +39,9 @@ def __init__(self, robot, arm, grab_entity):
         self.grab_entity_designator = grab_entity
 
         check_type(grab_entity, Entity)
+        assert self.robot.get_arm(
+            **self.REQUIRED_ARM_PROPERTIES), "None of the available arms meets all this class's" \
```
jlunenburg: Strictly speaking, I would even prefer `assert self.robot... is not None, ...`, to make it explicit that you're checking for a `None` return value rather than a `True` or `False`

robot_smach_states/src/robot_smach_states/util/designators/arm.py
```diff
@@ -13,6 +15,24 @@
 __author__ = 'loy'
 
 
+def const_resolve(arm_designator, additional_properties):
```
jlunenburg: You might consider calling this something like `resolve_extended`. I think that the `resolve` should always be const. This method is useful to resolve a designator with additional demands without changing the original. I'm curious about @LoyVanBeek 's view on this matter

robot_smach_states/src/robot_smach_states/util/designators/arm.py
```diff
@@ -13,6 +15,24 @@
...
+    """
```
jlunenburg: Check whitespaces (before `:param...`, after return)...

robot_smach_states/src/robot_smach_states/util/designators/arm.py
LoyVanBeek: I think this is not the way to do things. If a state requires a more specific arm than other states, that `ArmDesignator`, shared with those other states, should be extended. Otherwise, for example, state A has different, more lax requirements and ends up with a different arm than the more demanding state B, which are both part of the same process that should use the *same* arm in the end. 

Extending the arm requirements is OK, cloning designators to alter and potentially have the resolve differently is not.

jlunenburg: Thanks for your feedback. I guess that depends on where you define the requirements and how you implement it.

The usecase is thus the `ArmDesignator`. Suppose I build a hierarchical state machine: from the top level, I don't want to worry about the requirements I need to add. However, I do want 

1. a check at startup to make sure my robot has actually the required hardware to complete the task
2. at runtime, when I I resolve my designator, that my arm still meets the same requirement.

In the current implementation (the one in master), I need to add *all* requirements to the top-level designator. Even in case of a simple example, this is extremely unpractical (see, e.g., the arm designator in [this script](https://github.com/tue-robotics/tue_robocup/pull/1013/files#diff-1112c2269344b370678c16ef86befdfb).

The proposed implementation makes it therefore much easier for the user. If the user (at any level) wants to make sure a designator resolves to the same arm, it should be locked (that's what this method is for!). 

I actually consider it very bad practice if states are able to modify a designator (other than the `writing` that has been implemented). That makes it extremely hard to debug stuff: if you test a certain part of your code stand-alone it will behave different from a situation where it's integrated. 

This discussion, however, does bring up a flaw in the proposed design: what if a robot has two arms; one of them has a gripper, the other one knows what a 'prepare_grasp' trajectory is. In that case, it will pass the startup check but fail on runtime anyway. To prevent this, the level at which the arm check is performed should probably be the same as the level where it should be locked. For example, the arm designator in the `Grab` state should be locked throughout the entire state (otherwise, we'd better use a different method). Correspondingly, the `Grab` state should collect the requirements of all its child states, combine these and then perform the startup check.

LoyVanBeek: The locking is different than what you mention. That applies when e.g. I need an Entity on-top-of the table and I want to resolve the same ID after than object has been grasped off the table. That;s locking. 

In the case discussed here, I do think that modifying/extending/refining a designator (appending more requirements) is actually appropriate. By the end of the `machine.__init__` you know what arm  you really really want. And yes, if you use different states, you get different requirements. 

The order (the hierachy is a tree but the code of constructing that tree is still sequential) of when the requirements are added should not matter and I don't think it does given it's all dictionaries. 

We're in a pickle if things conflict though, but we can check for that. Eg. one state needs a suction cup and the other a parallel gripper etc.

jlunenburg: 'The locking is different': in what sense? If I position my two-armed robot to grasp using the left arm I definitely want to use the same arm for the next actions. I actually once implemented this by passing userdata and last time we used AMIGO we actually found out that the current implementation of the grab state *actually does not work anymore*. This indicates we *should* have a decent implementation that can either be locking or resolving just once and subsequently using a different mechanism.

Furthermore, I'm still not convinced on your argument on mutability: it still leads to different behavior while the same inputs are provided and as mentioned we're mixing the locking and arm requirements issues.

Maybe we should discuss this face to face to come to an 'optimal' design.

LoyVanBeek: > Maybe we should discuss this face to face to come to an 'optimal' design.

Yes, definitely. 

Locking is different is that the *same* designator can resolve to different things over time, potentially. Locking solves that, same value until it's unlocked. 

The issue here is that, under the hood, we copy a designator and suddenly we have multiple designators, that can each resolve to a different thing at the same moment. For the arms, we want all the states using something to resolve to the same arm as well. 
If we lock it but has the wrong value because of different requirements, locking is not the solution.

MatthijsBurgh:
@LarsJanssenTUe please resolve CI issues.

robot_smach_states/src/robot_smach_states/util/designators/arm.py
```diff
@@ -13,6 +15,24 @@
...
+    arm_designator_add_props = copy.copy(arm_designator)
+    for k, v in additional_properties.items():
+        if k in arm_designator_add_props.arm_properties:
+            for val in additional_properties[k]:
+                if val not in additional_properties[k]:
+                    arm_designator_add_props.arm_properties[k] += val
+        else:
+            arm_designator_add_props.arm_properties[k] = v
+    return arm_designator_add_props.resolve()
+
```
PetervDooren, line 35: Three things happen here and that is not clear from the name or docstring of this function.
1. You copy the designator (which, as stated above, interferes with locking behaviour)
2. You add requirements to your (copy of) arm designator
3. You resolve the copy and return it.
I would like to see a clear separation between these functions.

Also question. Where do you want to use this function?

PetervDooren (changes requested):

robot_smach_states/src/robot_smach_states/util/designators/arm.py
LarsJanssenTUe: @LoyVanBeek I am not far enough that I will be merging #1051 into this branch I and would like your input on Peters comment. Do you have some time Tuesday to help me think about this?

robot_smach_states/src/robot_smach_states/utility.py
```diff
@@ -357,22 +357,25 @@ class ResolveArm(smach.State):
     def __init__(self, arm, state_machine):
         """ Resolves, if possible, an arm for a state machine taking into account all the arm requirements
 
-        :param arm: given arm designator
+        :param arm: (lockable) arm designator
         :param state_machine: used state machine
         """
         smach.State.__init__(self, outcomes=['succeeded', 'failed'])
+
+        if not isinstance(arm, LockingDesignator):
```
MatthijsBurgh, line 365: I don't think this will work on the higher leve. As the original designator, `arm` is still used by the other states. So those aren't locked, but these should be! So we should check indeed that `arm` is a `LockingDesignator`, but we should raise an error. Not create a `LockingDesignator` ourselves

robot_smach_states/src/robot_smach_states/utility.py
```diff
@@ -357,22 +357,25 @@ class ResolveArm(smach.State):
...
         for k, v in arm_requirements.items():
-            if k in self.arm.arm_properties:
+            if k in self.arm.to_be_locked.arm_properties:
                 for val in arm_requirements[k]:
                     if val not in arm_requirements[k]:
-                        self.arm.arm_properties[k] += val
+                        self.arm.to_be_locked.arm_properties[k] += val
             else:
-                self.arm.arm_properties[k] = v
+                self.arm.to_be_locked.arm_properties[k] = v
```
MatthijsBurgh: Why the `to_be_locked`?

robot_smach_states/src/robot_smach_states/utility.py
```diff
@@ -357,22 +357,25 @@ class ResolveArm(smach.State):
     def __init__(self, arm, state_machine):
         """ Resolves, if possible, an arm for a state machine taking into account all the arm requirements
 
-        :param arm: given arm designator
+        :param arm: (lockable) arm designator
```
MatthijsBurgh, line 360: Not optional, should be enforced.

MatthijsBurgh (dismissed):

robot_smach_states/src/robot_smach_states/utility.py
LarsJanssenTUe: This is necessary to obtain the arm properties from the locking designator. This "to_be_locked" is defined in the locking designator.

MatthijsBurgh (approved):

PR #1017, opened 2020-03-08
added default arguments to Find state

PetervDooren:
merge together with https://github.com/tue-robotics/action_server/pull/132


robot_smach_states/src/robot_smach_states/navigation/find.py
```diff
@@ -94,8 +95,8 @@ class Find(smach.StateMachine):
     Find an entity based on a description. The description designator should
     have resolve type dict and it should contain at least a 'type' field
     """
-    def __init__(self, robot, knowledge, source_entity_designator, description_designator, area_name_designator,
-                 navigation_area_designator, found_entity_designator):
+    def __init__(self, robot, knowledge, source_entity_designator, description_designator, found_entity_designator,
+                 area_name="on_top_of", navigation_area_designator=""):
```
LarsJanssenTUe: I can understand the area_name default value, however the previous name suggests that it should be a designator and not a string, I assume you checked this? On top of that why would you give a default empty navigation_area_designator (again: designator?!)? I can't imagine that being a good idea in general.

robot_smach_states/src/robot_smach_states/navigation/find.py
```diff
@@ -105,11 +106,17 @@ def __init__(self, robot, knowledge, source_entity_designator, description_desig
                                    Inspect(robot=robot,
                                            entityDes=source_entity_designator,
                                            objectIDsDes=segmented_entities_designator,
-                                           searchArea=area_name_designator,
+                                           searchArea=area_name,
                                            navigation_area=navigation_area_designator),
-                                   transitions={'done': 'CHECK_IF_ENTITY_FOUND',
+                                   transitions={'done': 'WAIT',
```
LarsJanssenTUe: Please explain why we need to wait for a second, not only in this PR, but in the code to please.

robot_smach_states/src/robot_smach_states/navigation/find.py
PetervDooren: The previous names are deceptive as both area_name and navigation_area can be either strings of designators to strings.

The empty string is used to check whether this value was provided by the user much like you would use =None (which we could also use here). If it is empty we use a navigateToObserve

robot_smach_states/src/robot_smach_states/navigation/find.py
LarsJanssenTUe: I would still like to know why we have a seemingly random wait of half a second. If I see it like this I don't know what it is for and I would throw it out

robot_smach_states/src/robot_smach_states/navigation/find.py
LarsJanssenTUe: > If it is empty we use a navigateToObserve
Could you please put this in the docstring as well? That would make it really clear. 
Kudo's for adding the docstring by the way, this makes everything better. Nice job!

robot_smach_states/src/robot_smach_states/navigation/find.py
PetervDooren: We need to wait for some time in order to give ed time to process the updaterequest. I didn't time this exactly and it is likely not constant so it might be 0.1 second or 0.2 or 0.5.

robot_smach_states/src/robot_smach_states/navigation/find.py
PetervDooren: I think making it None is better for clarity.

robot_smach_states/src/robot_smach_states/navigation/find.py
```diff
@@ -94,8 +95,17 @@ class Find(smach.StateMachine):
...
+    def __init__(self, robot, knowledge, source_entity_designator, description_designator, found_entity_designator,
+                 area_name="on_top_of", navigation_area=None):
+        """
+        :param robot: Robot object
+        :param knowledge: Robocup knowledge object. Used to find an object by its properties
+        :param source_entity_designator: Designator resolving to EdEntity to inspect
+        :param description_designator: dict which contains at least a 'type' field
+        :param found_entity_designator: Designator the found entity can be written to
+        :param area_name: str or str Designator describing the area of the source entity to inspect
+        :param navigation_area: str or str Designator describing the area to drive to for the inspect
```
LarsJanssenTUe:
```suggestion
        :param navigation_area: str or str Designator describing the area to drive to for the inspect. The default, 'None', will result in a navigateToObserve otherwise a navigateToSymbolic will be used.
```
This assumes that otherwise a navigateToSymbolic is used

robot_smach_states/src/robot_smach_states/navigation/find.py
LarsJanssenTUe: Shouldn't this be handled in the inspect itself? I would agree with adding a sleep there, but not here.

alberth:
Waiting some random time assumes it is always long enough, even if in the future, we may change eg ed to become slower (intentionally or by accident). As such it is a highly fragile solution.
You should wait by blocking on a response from ed. That the only way you know ed has done its thing for sure.

robot_smach_states/src/robot_smach_states/navigation/find.py
PetervDooren: I'll create an issue for this and revert these changes as this goes outside of the scope of this PR

PR #1023, opened 2020-03-14
Move arm handover

alberth:
First step of arm splitting, move the handover function to its own class.

### DIFF
--- a/robot_skills/src/robot_skills/arms.py
+++ b/robot_skills/src/robot_skills/arms.py
@@ -330,6 +330,58 @@ class GripperState(object):
     CLOSE = "close"
 
 
+class ArmHandover(object):
+    """
+    :var robot_name: Name of the robot.
+    :var arm_name: Name of the arm (typically 'left' or 'right').
+    """
+    def __init__(self, robot_name, arm_name):
+        self.robot_name = robot_name
+        self.arm_name = arm_name
+
+    def handover_to_human(self, timeout=10):
+        """
+        Handover an item from the gripper to a human.
+
+        Feels if user slightly pulls or pushes (the item in) the arm. On timeout, it will return False.
+        :param timeout: timeout in seconds
+        :return: True or False
+        """
+        pub = rospy.Publisher('/'+self.robot_name+'/handoverdetector_'+self.arm_name+'/toggle_robot2human',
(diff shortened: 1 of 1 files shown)


robot_skills/src/robot_skills/arms.py
```diff
@@ -330,6 +330,59 @@ class GripperState(object):
...
+    Handover part of the arm functionality.
+
+    :var robot_name: Name of the robot.
+    :var arm_name: Name of the arm (typically 'left' or 'right').
+    """
+    def __init__(self, robot_name, arm_name):
+        self.robot_name = robot_name
+        self.arm_name = arm_name
+
+    def handover_to_human(self, timeout=10):
```
LarsJanssenTUe: The definitions handover_to_human and handover_to_robot are now defined twice (here and in the arm class). I think they can be removed here if the functionality (return exchange_with_human) is shifted to the arm class.

robot_skills/src/robot_skills/arms.py
```diff
@@ -330,6 +330,59 @@ class GripperState(object):
...
+        """
+        Handover an item from a human to the robot.
+
+        Feels if user slightly pushes an item in the gripper. On timeout, it will return False.
+        :param timeout: timeout in seconds
+        :return: True or False
+        """
+        return self._exchange_with_human(True, timeout)
+
+    def _exchange_with_human(self, to_robot, timeout=10):
```
LarsJanssenTUe: Good idea to combine this! I like it!

robot_skills/src/robot_skills/arms.py
alberth: Actually, this and #1024 are the start of cleaning out the much of the Arm class by decomposing its code into a set of smaller interfaces. this and #1024 are the first steps in that.

However, since that is a complicated operation on a crucial part of the robot, I am taking baby steps, using knowledge gained in #1011. As such, your suggestion is in the opposite direction of what is being aimed for :)

LarsJanssenTUe: I like that you are taking small steps and I think it is not a big issue to have a little bit of duplicate code, but I guess that in the end we want to get rid of this. But for now I guess this is fine.

PR #1024, opened 2020-03-14
Create functions for grasp compute

alberth:
Deduplicate assigning position and orientation in the grasp computation.

### DIFF
--- a/robot_skills/src/robot_skills/arms.py
+++ b/robot_skills/src/robot_skills/arms.py
@@ -521,15 +521,15 @@ class Arm(RobotPart):
 
         grasp_precompute_goal.allowed_touch_objects = allowed_touch_objects
 
-        set_grasp_goal_position(grasp_precompute_goal.goal, frame_in_baselink.p)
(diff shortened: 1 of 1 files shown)


alberth:
Actually, it is not tested and I don't know how to do that, as in all these years, nobody has managed to explain to me what I need to install to get the simulator running without magic `tue`-_something_ command.
We can have another attempt at that.

LoyVanBeek:
> Actually, it is not tested and I don't know how to do that, as in all these years, nobody has managed to explain to me what I need to install to get the simulator running without magic `tue-something` command.

@MatthijsBurgh can you help @alberth with this?

LarsJanssenTUe:
@alberth @MatthijsBurgh I assume that @alberth does not have a simulator yet right? @alberth I could test these PR's for you, but that does not solve anything in the long run, so I suggest you get the simulator. I know that it is possible to do it without running the 'magic tue-something command', however this would probably be a pain (hence the reason for this magical command), so my question would be: why not use this?

PR #1028, opened 2020-03-27
separate constraint functions from states

PetervDooren:
From the creators of "that time we tried to do this using Designators #1006" comes the same thing without the designators. Arguably this is a more elegant solution as it involves a lot less inheritance.

### DIFF
--- a/robot_smach_states/examples/navigation/constraint_functions/example_arms_reach_constraints.py
+++ b/robot_smach_states/examples/navigation/constraint_functions/example_arms_reach_constraints.py
@@ -47,7 +47,6 @@ if __name__ == "__main__":
     entity_designator = Designator(entity)
 
     rospy.loginfo("When you have an entity rather than a pose, consider wrapping it in an attribute designator")
-    rospy.loginfo("When you have an entity rather than a pose, consider wrapping it in an attribute designator")
     rospy.loginfo("Running: pc, oc = arms_reach_constraint(AttrDesignator(entity_designator, 'pose', resolve_type=FrameStamped), arm_designator)")
     pc, oc = arms_reach_constraint(AttrDesignator(entity_designator, 'pose', resolve_type=FrameStamped), arm_designator)
     rospy.loginfo("pc becomes : {}".format(pc))
--- a/robot_smach_states/examples/navigation/constraint_functions/example_arms_reach_constraints.py
+++ b/robot_smach_states/examples/navigation/constraint_functions/example_arms_reach_constraints.py
@@ -1,0 +1,54 @@
+import rospy
+import argparse
+
+from robot_skills.get_robot import get_robot
+from robot_skills.util.entity import Entity
+from robot_skills.util.kdl_conversions import frame_stamped, FrameStamped
+# Robot Smach States
+from robot_smach_states.navigation.constraint_functions.arms_reach_constraints import arms_reach_constraint
+from robot_smach_states.util.designators.arm import ArmDesignator
+from robot_smach_states.util.designators.core import Designator
+from robot_smach_states.util.designators.utility import AttrDesignator
+
+if __name__ == "__main__":
+    """
+    Example demonstrating how to use the arms_reach_constraint function to generate navigation constraints
+    Make sure you have a (robot)-start running since a robot object is created for this.
+    """
+
+    parser = argparse.ArgumentParser(description="Example arms_reach_constaint() function")
+    parser.add_argument("--robot", default="hero", help="Robot name (amigo, hero, sergio)")
+    args = parser.parse_args()
+
+    rospy.init_node("example_arms_reach_constraints")
+    robot = get_robot(args.robot)
+
+    # get an armdesignator and a pose designator
+    arm_designator = ArmDesignator(robot, {}, name="arm_designator")
+    pose = frame_stamped("/map", 0, 0, 1.3)
+    pose_designator = Designator(pose, name="frame designator")
+
+    # basic functionality
+    rospy.loginfo("Example of the basic funtionality")
+    rospy.loginfo("Running: pc, oc = arms_reach_constraint(pose_designator, arm_designator)")
+    pc, oc = arms_reach_constraint(pose_designator, arm_designator)
+    rospy.loginfo("pc becomes : {}".format(pc))
+    rospy.loginfo("oc becomes : {}".format(oc))
+
+    # dont look at the frame
+    rospy.loginfo("Example where we don't require the robot to look at the pose")
+    rospy.loginfo("Running: pc, oc = arms_reach_constraint(pose_designator, arm_designator, look=False)")
+    pc, oc = arms_reach_constraint(pose_designator, arm_designator, look=False)
+    rospy.loginfo("pc becomes : {}".format(pc))
+    rospy.loginfo("oc becomes : {}".format(oc))
+
+    # Use an entity
+    entity = Entity("dummy_entity", "dummy_type", "/map", pose.frame, None, None, None, None)
+    entity_designator = Designator(entity)
+
+    rospy.loginfo("When you have an entity rather than a pose, consider wrapping it in an attribute designator")
+    rospy.loginfo("When you have an entity rather than a pose, consider wrapping it in an attribute designator")
+    rospy.loginfo("Running: pc, oc = arms_reach_constraint(AttrDesignator(entity_designator, 'pose', resolve_type=FrameStamped), arm_designator)")
+    pc, oc = arms_reach_constraint(AttrDesignator(entity_designator, 'pose', resolve_type=FrameStamped), arm_designator)
+    rospy.loginfo("pc becomes : {}".format(pc))
+    rospy.loginfo("oc becomes : {}".format(oc))
--- a/robot_smach_states/examples/navigation/constraint_functions/example_symbolic_constraints.py
+++ b/robot_smach_states/examples/navigation/constraint_functions/example_symbolic_constraints.py
@@ -1,0 +1,29 @@
+import rospy
+import argparse
+
+from robot_skills.get_robot import get_robot
+# Robot Smach States
+from robot_smach_states.navigation.constraint_functions.symbolic_constraints import symbolic_constraint
+from robot_smach_states.util.designators.ed_designators import EntityByIdDesignator
+
+if __name__ == "__main__":
+    """
+    Example demonstrating how to use the symbolic_constraint() function
+    Make sure you have both (robot)-start and ED running.
+    """
+
+    parser = argparse.ArgumentParser(description="Example symbolic_constaint() function")
+    parser.add_argument("--robot", default="hero", help="Robot name (amigo, hero, sergio)")
+    args = parser.parse_args()
+
+    rospy.init_node("example_symbolic_constraints")
+    robot = get_robot(args.robot)
+
+    # get an entitydesignator
+    entity = EntityByIdDesignator(robot, 'dinner_table', name='entity_designator')
+
+    # with rotation
+    rospy.loginfo("symbolic constraint corresponding to 'in_front_of' the 'dinner_table':")
+    pc, oc = symbolic_constraint(robot, {entity: "in_front_of"})
+    rospy.loginfo("pc becomes : {}".format(pc))
+    rospy.loginfo("oc becomes : {}".format(oc))
--- a/robot_smach_states/examples/navigation/constraint_functions/example_symbolic_constraints.py
+++ b/robot_smach_states/examples/navigation/constraint_functions/example_symbolic_constraints.py
@@ -3,7 +3,7 @@ import argparse
 
 from robot_skills.get_robot import get_robot
 # Robot Smach States
-from robot_smach_states.navigation.constraint_functions.symbolic_constraints import symbolic_constraint
+from robot_smach_states.navigation.constraint_functions.symbolic_constraints import symbolic_constraint, room_constraint
 from robot_smach_states.util.designators.ed_designators import EntityByIdDesignator
 
 if __name__ == "__main__":
@@ -22,8 +22,17 @@ if __name__ == "__main__":
     # get an entitydesignator
     entity = EntityByIdDesignator(robot, 'dinner_table', name='entity_designator')
 
-    # with rotation
+    # in front of the dinner table
     rospy.loginfo("symbolic constraint corresponding to 'in_front_of' the 'dinner_table':")
     pc, oc = symbolic_constraint(robot, {entity: "in_front_of"})
+    rospy.loginfo("pc becomes : {}".format(pc))
+    rospy.loginfo("oc becomes : {}".format(oc))
+
+    # get a roomdesignator
+    room = EntityByIdDesignator(robot, 'kitchen', name='room_designator')
+
+    # room constraint
+    rospy.loginfo("room constraint corresponding to 'kitchen':")
+    pc, oc = room_constraint(robot, room)
     rospy.loginfo("pc becomes : {}".format(pc))
     rospy.loginfo("oc becomes : {}".format(oc))
--- a/robot_smach_states/src/robot_smach_states/navigation/constraint_functions/arms_reach_constraints.py
+++ b/robot_smach_states/src/robot_smach_states/navigation/constraint_functions/arms_reach_constraints.py
@@ -14,9 +14,9 @@ from cb_planner_msgs_srvs.msg import OrientationConstraint, PositionConstraint
 def arms_reach_constraint(pose_designator, arm_designator, look=True):
     """
     Position so that the arm can reach the position/entity
-    :param pose_designator: designator that resolves to a FrameStamped
-    :param arm_designator: which arm to use for manipulation
-    :param look: whether or not the orientation must be constrained as well
+    :param pose_designator: designator that resolves to a FrameStamped of the point to be reached
+    :param arm_designator: PublicArmDesignator, arm to use for manipulation
+    :param look: bool, whether or not the orientation must be constrained as well
     :return: tuple(PositionConstraint, OrientationConstraint)
     """
     arm = arm_designator.resolve()
--- a/robot_smach_states/src/robot_smach_states/navigation/constraint_functions/arms_reach_constraints.py
+++ b/robot_smach_states/src/robot_smach_states/navigation/constraint_functions/arms_reach_constraints.py
@@ -22,14 +22,14 @@ def arms_reach_constraint(pose_designator, arm_designator, look=True):
     """
     arm = arm_designator.resolve()
     if not arm:
-        rospy.logerr("Could not resolve arm")
+        rospy.logerr("Could not resolve arm, Designator {} did not resolve".format(arm_designator))
         return None
 
     radius = math.hypot(arm.base_offset.x(), arm.base_offset.y())
 
     pose = pose_designator.resolve()
     if not pose:
-        rospy.logerr("No such place_pose")
+        rospy.logerr("No such place_pose, Designator {} did not resolve".format(pose_designator))
         return None
 
     try:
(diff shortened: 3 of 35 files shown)


PetervDooren:
@jlunenburg, what do you think of this approach compared to the one using designators? Is this more useful in practice?

PetervDooren:
It's alright to just skim through. This is just a proof of concept.

robot_smach_states/src/robot_smach_states/navigation/constraint_functions/arms_reach_constraints.py
```diff
@@ -0,0 +1,51 @@
...
+    arm = arm_designator.resolve()
+    if not arm:
+        rospy.logerr("Could not resolve arm")
+        return None
+
+    radius = math.hypot(arm.base_offset.x(), arm.base_offset.y())
+
+    pose = pose_designator.resolve()
+    if not pose:
+        rospy.logerr("No such place_pose")
```
LoyVanBeek: Plz log which designator failed, in each of them. You should be able to just `str()` them

PetervDooren: Currently this is a proof of concept. I want to decide which approach is better, passing functions or using designators. After that decision is made I will further develop either this PR or #1006.

LoyVanBeek:
I like it! Makes things much easier to test offline. 
I also like the idea of #1006 but this current PR is a more obvious (a huge positive for reviewing) and less rigorous change. It could be a step towards the concept behind #1006 if we still want that at a later time. 
I'd say: finalize and merge this, add a metric shit-ton of tests and then consider if we want to continue with the concept of #1006.

PetervDooren:
> I like it! Makes things much easier to test offline.
> I also like the idea of #1006 but this current PR is a more obvious (a huge positive for reviewing) and less rigorous change. It could be a step towards the concept behind #1006 if we still want that at a later time.
> I'd say: finalize and merge this, add a metric shit-ton of tests and then consider if we want to continue with the concept of #1006.

I'd like  to do this only once.  So if we want the concept of #1006 anyway we might as well do that now.

LoyVanBeek:
Then:
"finalize and merge this, add a metric shit-ton of tests ~and then consider if we want to continue with the concept of #1006.~"

robot_smach_states/src/robot_smach_states/navigation/constraint_functions/arms_reach_constraints.py
```diff
@@ -0,0 +1,51 @@
...
+from geometry_msgs.msg import Point
+
+# TU/e Robotics
+from cb_planner_msgs_srvs.msg import OrientationConstraint, PositionConstraint
+
+
+def arms_reach_constraint(pose_designator, arm_designator, look=True):
+    """
+    Position so that the arm can reach the position/entity
+    :param pose_designator: designator that resolves to a FrameStamped
```
alberth: An empty line between the description and the lines that start with : is required for the doc processor afaik.

alberth:
I like this approach. unlike in #1006, the designators stay simple and independent of what is being resolved. Also, the functions split away constraint solving from the other code, and you get flexibility by being able to write new functions as needed.

robot_smach_states/src/robot_smach_states/navigation/constraint_functions/arms_reach_constraints.py
PetervDooren: Done

robot_smach_states/examples/navigation/constraint_functions/example_arms_reach_constraints.py
```diff
@@ -0,0 +1,54 @@
...
+    pc, oc = arms_reach_constraint(pose_designator, arm_designator, look=False)
+    rospy.loginfo("pc becomes : {}".format(pc))
+    rospy.loginfo("oc becomes : {}".format(oc))
+
+    # Use an entity
+    entity = Entity("dummy_entity", "dummy_type", "/map", pose.frame, None, None, None, None)
+    entity_designator = Designator(entity)
+
+    rospy.loginfo("When you have an entity rather than a pose, consider wrapping it in an attribute designator")
+    rospy.loginfo("When you have an entity rather than a pose, consider wrapping it in an attribute designator")
```
LoyVanBeek, line 50: Twice?

robot_smach_states/examples/navigation/constraint_functions/example_symbolic_constraints.py
```diff
@@ -0,0 +1,29 @@
...
+
+    rospy.init_node("example_symbolic_constraints")
+    robot = get_robot(args.robot)
+
+    # get an entitydesignator
+    entity = EntityByIdDesignator(robot, 'dinner_table', name='entity_designator')
+
+    # with rotation
+    rospy.loginfo("symbolic constraint corresponding to 'in_front_of' the 'dinner_table':")
+    pc, oc = symbolic_constraint(robot, {entity: "in_front_of"})
```
LoyVanBeek, line 27: It might be good to expand this a little and do a pose & room for example.

robot_smach_states/test/navigation/constraint_functions/test_arms_reach_constraints.py
```diff
@@ -0,0 +1,35 @@
...
+        self.assertIsNotNone(oc)
+
+    def test_no_look(self):
+        robot = Mockbot()
+        arm = ArmDesignator(robot, {}, name="arm_designator")
+        frame = Designator(frame_stamped("/map", 0, 0, 0), name="frame_designator")
+
+        pc, oc = arms_reach_constraint(frame, arm, look=False)
+
+        self.assertIsNotNone(pc)
```
LoyVanBeek, line 30: What I'd  ke to see for all of these tests is to check the content of the `pc`, `oc` variables. 
The constraints are expressed at mathematical equations in a simple language, as a `str`. I'd like to be sure that the functions output the `str`s are correct. 

That holds for all of these tests.

robot_smach_states/test/navigation/constraint_functions/test_compound_constraints.py
```diff
@@ -0,0 +1,45 @@
...
+                                      lambda: dummy_constraint("3")])
+        pc1, oc1 = dummy_constraint("1")
+        pc2, oc2 = dummy_constraint("2")
+        pc3, oc3 = dummy_constraint("3")
+
+        self.assertIn(pc1.constraint, pc.constraint)
+        self.assertIn(pc2.constraint, pc.constraint)
+        self.assertIn(pc3.constraint, pc.constraint)
+
+        # only one orientation constraint is allowed under current system.
```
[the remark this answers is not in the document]
PetervDooren: It's the first one in the list. Which basically means that you can only specify one orientationconstraint. This is a limitation of the orientationconstraint itself and improving it is IMO out of the scope of this PR

robot_smach_states/examples/navigation/constraint_functions/example_arms_reach_constraints.py
PetervDooren: Oops

robot_smach_states/examples/navigation/constraint_functions/example_symbolic_constraints.py
PetervDooren: I added an example for room_constraints, not sure what you mean by pose though

robot_smach_states/test/navigation/constraint_functions/test_arms_reach_constraints.py
PetervDooren: The tricky part is that we don't want to overly specify the format in which the string is constructed. For example "x>0" and "x > 0" are the same mathematical expressions but are different strings.

PetervDooren: I added as many checks as I found reasonable. But that is still not enough to verify the mathematical expressions are correct due to my previous point

robot_smach_states/examples/navigation/constraint_functions/example_symbolic_constraints.py
LoyVanBeek: Sorry, I meant 'entity and room', eg. a dictionary with multiple entries.

PetervDooren:
@LoyVanBeek I would like to write the tests in this PR. 
I just didn't know how to make the tests more elaborate without hindering inconsequential changes.
I should probably have a look at cb_base navigation and see how the position constraints should be specified.  Then I can make these tests more elaborate. But if any of you have suggestions at this point I would be more than happy to accomdate these

robot_smach_states/test/navigation/constraint_functions/test_arms_reach_constraints.py
LoyVanBeek: That is a good point. We could strip all whitespace of course but what matters is the underlying expression tree. 
For now, I'd be quite happy if we can get a few of such tests in. Just execute a test, record the result and check if it makes sense given the input. Then use that as the expected result. Then, we'll at least know when we break something.

robot_smach_states/src/robot_smach_states/navigation/navigate_to_grasp.py
```diff
@@ -1,72 +1,38 @@
...
-        # Outer radius
-        ro = "(x-%f)^2+(y-%f)^2 < %f^2"%(x, y, radius+0.075)
-        ri = "(x-%f)^2+(y-%f)^2 > %f^2"%(x, y, radius-0.075)
-        pc = PositionConstraint(constraint=ri+" and "+ro, frame="/map")
-        oc = OrientationConstraint(look_at=Point(x, y, 0.0), frame="/map", angle_offset=angle_offset)
+            arm_designator = UnoccupiedArmDesignator(self.robot, {})
 
-        return pc, oc
+        super(NavigateToGrasp, self).__init__(robot,
+                                              lambda: arms_reach_constraint(pose_designator, arm_designator, look=True),
```
alberth, line 37: If you assign this lambda to a variable above line, this line becomes much shorter.

robot_smach_states/src/robot_smach_states/navigation/navigate_to_waypoint.py
```diff
@@ -1,59 +1,31 @@
...
-            oc = OrientationConstraint(look_at=Point(x+10, y, 0.0), angle_offset=rz, frame="/map")
-
-        return pc, oc
+    """
+    Navigate to a waypoint in the world model
+
+    :param waypoint_designator: designator resolving to the waypoint
+    :param radius (default 0.15): allowed distance to the waypoint
+    :param look_at_designator: Entity the robot should face from the waypoint. If not specified the robot will adopt
+     the orientation of the waypoint
```
alberth, line 18: At other places you indent this line (which I also prefer, as it makes the continuation more recognizable).

robot_smach_states/src/robot_smach_states/navigation/navigation.py
```diff
@@ -219,7 +229,7 @@ def __init__(self, robot, reset_head=True, speak=True, reset_pose=True, input_ke
 
             with sm_nav:
 
-                smach.StateMachine.add('GET_PLAN',                          getPlan(self.robot, self.generateConstraint, self.speak),
+                smach.StateMachine.add('GET_PLAN',                          getPlan(self.robot, constraint_function, self.speak),
                     transitions={'unreachable'                          :   'unreachable',
                                  'goal_not_defined'                     :   'goal_not_defined',
                                  'goal_ok'                              :   'EXECUTE_PLAN'})
```
alberth, line 235: Not sure why people add such large amounts of white space, but it existed already

robot_smach_states/test/navigation/constraint_functions/test_pose_constraints.py
```diff
@@ -0,0 +1,47 @@
...
+
+        # verify positionconstraint
+        self.assertIn("x", pc.constraint)
+        self.assertIn("y", pc.constraint)
+        self.assertEqual(pc.frame, frame_id)
+
+        # verify orientationconstraint
+        self.assertIsNone(oc)
+
+    def test_orientation(self):
```
alberth, line 27: Could perhaps rename to `test_with_orientation`, which better expresses the opposite of the previous test.

robot_smach_states/src/robot_smach_states/navigation/constraint_functions/symbolic_constraints.py
```diff
@@ -0,0 +1,46 @@
...
+    :return: (tuple(PositionConstraint, OrientationConstraint)). If one of the entities does not resolve,
+        None is returned.
+    """
+    entity_id_area_name_map = {}
+    for desig, area_name in entity_designator_area_name_map.iteritems():
+        entity = desig.resolve()
+        try:
+            area_name = area_name.resolve()
+        except:
+            pass
```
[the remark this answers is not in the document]
PetervDooren: This is an artefact from the previous code. It was meant to make it so that you could either use a designator or a value (which is a bad practice but not the focus of this PR). I made it less ugly by checking for a "resolve_type" attribute

robot_smach_states/src/robot_smach_states/navigation/constraint_functions/radius_constraints.py
```diff
@@ -0,0 +1,64 @@
...
+
+    if len(ch) > 0:  # If a convex hull is present, use this to create the position constraint
+        pci = ""  # Create an empty position constraint
+
+        for i in xrange(len(ch)):  # Iterate over the convex hull
+            j = (i + 1) % len(ch)
+            dx = ch[j].x() - ch[i].x()
+            dy = ch[j].y() - ch[i].y()
+
+            length = (dx * dx + dy * dy) ** .5
```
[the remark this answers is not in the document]
PetervDooren: Used math.hypot instead

robot_smach_states/src/robot_smach_states/navigation/constraint_functions/arms_reach_constraints.py
```diff
@@ -0,0 +1,51 @@
...
+    :param arm_designator: which arm to use for manipulation
+    :param look: whether or not the orientation must be constrained as well
+    :return: tuple(PositionConstraint, OrientationConstraint)
+    """
+    arm = arm_designator.resolve()
+    if not arm:
+        rospy.logerr("Could not resolve arm")
+        return None
```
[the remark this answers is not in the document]
LoyVanBeek: The type contract of a Designator is actually that: if it can't find an answer, it returns `None`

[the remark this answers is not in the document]
PetervDooren: I think Albert is referring to the function itself, which returns None here rather than a tuple of constraints. I already adjusted the type contract to reflect this

robot_smach_states/test/navigation/constraint_functions/test_waypoint_constraints.py
```diff
@@ -37,20 +44,26 @@ def test_no_look(self):
...
+        pc, oc = waypoint_constraint(entity, radius, look=False)
 
         # verify positionconstraint
-        self.assertIn("x", pc.constraint)
-        self.assertIn("y", pc.constraint)
+        constraint_string = pc.constraint
+        constraint_string = constraint_string.replace("^", "**")
         self.assertEqual(pc.frame, frame_id)
 
+        verification_string = "(x-{})**2+(y-{})**2 < {}**2".format(x_coordinate, y_coordinate, radius)
```
LoyVanBeek, line 62: Nice! 

Maybe it's handy to put code to check the constraint in a separate function that is given the verification_string and optionally a specification of the range we want to check it over (those linspaces). 
I'd then pass that verification_string in the same format as the original, so do the `^` to `**` on both. 

Then here you can just call `self.assertTrue(constraint_outcomes_equal(constraint_string, verification_string))` and it's super clear what is going on.
In that `constraint_outcomes_equal` (or a better name) you'll haveto document all the reasons why this is the way it's done and not eg. the string comparison I proposed and the reasons why that wouldn't work. Because that question will pop up for other reviewers and later.

PetervDooren: Seperated the comparison function. However I am not sure that comparing a matrix of evenly spaced points is the best way to go about this. Because right now we test points that are 10cm apart, which is a lot for some position constraints. I think it might be better to test a handful of strategically chosen points rather than a coarse array of effectively random points

robot_smach_states/src/robot_smach_states/navigation/navigation.py
PetervDooren: Now is as good a time as any to clean it up.

robot_smach_states/src/robot_smach_states/navigation/constraint_functions/arms_reach_constraints.py
```diff
@@ -0,0 +1,51 @@
...
+from cb_planner_msgs_srvs.msg import OrientationConstraint, PositionConstraint
+
+
+def arms_reach_constraint(pose_designator, arm_designator, look=True):
+    """
+    Position so that the arm can reach the position/entity
+    :param pose_designator: designator that resolves to a FrameStamped
+    :param arm_designator: which arm to use for manipulation
+    :param look: whether or not the orientation must be constrained as well
+    :return: tuple(PositionConstraint, OrientationConstraint)
```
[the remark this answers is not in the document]
PetervDooren: formatted the code. Not for python3 preparation however

robot_smach_states/src/robot_smach_states/navigation/constraint_functions/arms_reach_constraints.py
```diff
@@ -18,7 +18,8 @@ def arms_reach_constraint(pose_designator, arm_designator, look=True):
     :param pose_designator: designator that resolves to a FrameStamped of the point to be reached
     :param arm_designator: PublicArmDesignator, arm to use for manipulation
     :param look: bool, whether or not the orientation must be constrained as well
-    :return: tuple(PositionConstraint, OrientationConstraint), if a designator does not resolve None is returned
+    :return: navigation constraints, if a designator does not resolve None is returned
+    :type: tuple(PositionConstraint, OrientationConstraint)
```
LoyVanBeek, line 22: You can specify the return type with `:rtype`

robot_smach_states/test/navigation/constraint_functions/util.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def constraint_strings_equal(constraint, verification):
+    """
+    check that two constraint strings lead to the same result.
+    We cannot simply do string comparison as we want to verify that the underlying mathematical expression is identical
+     rather than the exact string
+    :param constraint: string containing the first constraint
+    :param verification: string containing the second constraint
+    :return: whether or not the constraints are equal.
```
LoyVanBeek, line 11: I'd add the result tuple as the `rtype`.

robot_smach_states/test/navigation/constraint_functions/test_symbolic_constraints.py
```diff
@@ -26,6 +26,24 @@ def test_base(self):
...
+        area_name1 = "area"
+
+        dummy_id_2 = "dummy"
+        e2 = Entity(dummy_id_2, "dummy_type", "/map", None, None, None, None, None)
+        entity2 = Designator(e2, name="entity designator")
+        area_name2 = "area"
+
+        pc, oc = symbolic_constraint(self.robot, {entity1: area_name1, entity2: area_name2})
+
+        self.assertIsNotNone(pc)
```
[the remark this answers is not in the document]
PetervDooren: Symbolic constraints can only specify position constraints. So we must assert that no orientation constraint is created by accident.
I added a small comment but I think this should be clear if someone understands what the symbolic constraints function is for. (If not they should not be working on the test)

robot_smach_states/test/navigation/constraint_functions/test_waypoint_constraints.py
LoyVanBeek: We could consider https://stackoverflow.com/questions/37112738/sympy-comparing-expressions but that's overkill I guess. 
Going down to comparing at a higher resolution gives more details of course, but I think these unit tests should mostly tell if nothing broke really. If that case, the outlines of the functions will be wildly different I expect.

PR #1031, opened 2020-04-18
(robot_skills) small doc fixes

MatthijsBurgh:


robot_skills/src/robot_skills/sergio.py
```diff
@@ -4,8 +4,15 @@
...
 class Sergio(robot.Robot):
-    """docstring for Sergio"""
+    """
+    Sergio
+    """
     def __init__(self, wait_services=False):
+        """
+        Constructor
+
+        :param wait_services: Not supported anymore by robot class
```
LoyVanBeek, line 14: Can we issue a deprecation warning when this is used?

robot_skills/src/robot_skills/robot.py
```diff
@@ -22,6 +22,16 @@ class Robot(object):
...
     def __init__(self, robot_name="", wait_services=False, tf_listener=None):
+        """
+        Constructor
+
+        :param robot_name: Name of the robot
+        :type robot_name: str
+        :param wait_services: Not used anymore
+        :type wait_services: bool
+        :param tf_listener: tf_listener object
+        :type tf_listener: Union[tf.TransformListener, None]
```
LoyVanBeek, line 33: IIRC, there is also `Optional[tf.TransformListener]` which would also be a bit clearer IMO

robot_skills/src/robot_skills/sergio.py
MatthijsBurgh: We can provide a deprecation warning if the value is `True`. But it needs a lot of refactoring to determine if it set by the user or not in case of `False`.

LoyVanBeek: OK, it's a start. There's only a few cases where we use it anyways I expect

PR #1033, opened 2020-05-05
Locking bug

MatthijsBurgh:
@LoyVanBeek although this is pretty old. I still think this one commit should be merged. As right now it doesn't return in that one special case.


robot_smach_states/src/robot_smach_states/util/designators/ed_designators.py
```diff
@@ -526,6 +526,7 @@ def _resolve(self):
                 entity = self.to_be_locked.resolve()  # Then find  out what we should remember
                 if entity:  # If we can find what to remember
                     self._locked_to_id = entity.id  # remember its ID.
+                    return entity
```
alberth, line 529: Bit late to the party, but shouldn't it return something if `if entity` fails?

PR #1038, opened 2020-06-16
fix example grasping

PetervDooren:


robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -103,17 +103,7 @@ def execute(self, userdata=None):
...
 
-        try:
-            # Transform to base link frame
-            goal_bl = goal_map.projectToFrame(self.robot.robot_name+'/base_link', tf_listener=self.robot.tf_listener)
-            if goal_bl is None:
-                rospy.logerr('Transformation of goal to base failed')
-                return 'failed'
-        except tf2_ros.TransformException as tfe:
-            rospy.logerr('Transformation of goal to base failed: {0}'.format(tfe))
-            return 'failed'
```
LarsJanssenTUe, line 116: The reason for removing this is that this behavior acts limiting. Down the line the same is checked, however with the option of another outcome, which we need in this particular case.

LarsJanssenTUe:
Apparently this can be done without the changes to grab.py as is done in #1039

robot_smach_states/examples/manipulation/example_grasping.py
```diff
@@ -33,9 +36,10 @@
     pose = FrameStamped(frame=kdl.Frame(kdl.Rotation.RPY(0.0, 0.0, 0.0), kdl.Vector(args.x, args.y, args.z)),
                         frame_id="/map")
 
-    robot.ed.update_entity(id=entity_id, frame_stamped=pose)
+    shape = RightPrism([kdl.Vector(0, 0, 0), kdl.Vector(0, 0.05, 0), kdl.Vector(0.05, 0.05, 0), kdl.Vector(0.05, 0, 0)], -0.1, 0.1)
+    item = Entity("dummy_id", "dummy_type", pose.frame_id, pose.frame, shape, None, None, None)
 
-    item = ds.EdEntityDesignator(robot, id=entity_id)
+    item = ds.Designator(item)
```
LoyVanBeek, line 42: It might be use to mention here as documentation that the `resolve_type` of the `Designator` must be `Entity` for `Grab` to accept it.

PR #1050, opened 2020-10-06
empty the arm when reset is called.

PetervDooren:


MatthijsBurgh (changes requested):
This doesn't match the docstring.

LarsJanssenTUe (dismissed):

robot_skills/src/robot_skills/arms.py
```diff
@@ -618,7 +618,7 @@ def send_joint_trajectory(self, configuration, timeout=5.0, max_joint_vel=0.7):
 
     def reset(self, timeout=0.0):
         """
-        Put the arm into the 'reset' pose
+        Put the arm into the 'reset' pose. And open the gripper
```
MatthijsBurgh, line 621: How about setting the occupied_by to None?

MatthijsBurgh (changes requested):

robot_skills/src/robot_skills/arms.py
PetervDooren: Opening the gripper generally results in that.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -184,7 +184,6 @@ def execute(self, userdata=None):
             self.robot.speech.speak('I am sorry but I cannot move my arm to the object position', block=False)
             rospy.logerr('Grasp failed')
             arm.reset()
-            arm.send_gripper_goal('close', timeout=0.0)
```
LarsJanssenTUe, line 187: First this line was redundant, however now, since the reset function is changed, removing this line changes the behaviour of the state. Since if this if-statement holds a 'failed' is returned, which leads to the 'resetonFailure' state, in which this different behaviour seems to work out fine.

robot_smach_states/src/robot_smach_states/manipulation/place.py
```diff
@@ -178,9 +178,6 @@ def execute(self, userdata=None):
             rospy.logwarn('Retraction failed')
             arm.cancel_goals()
 
-        # Close gripper
-        arm.send_gripper_goal('close', timeout=0.0)
```
LarsJanssenTUe, line 182: First this line, was redundant, now it is not due to a change in the reset function. Therefore the state of the gripper will be different at the end of the put state. This is not a problem in itself, but might change the behaviour of the robot in the current implementation of certain challenges. Again not a problem in itself, but potentially undesirable.

LarsJanssenTUe (approved):
Can change the behaviour of the robot in unexpected places, but the required changes would be trivial. The plus side of merging this is the removal of the occupied by variable upon resetting. I say the benefits outweigh the costs

robot_smach_states/src/robot_smach_states/manipulation/place.py
PetervDooren: Yes this may change behaviour in challenges. But I would argue that the responsibility lies with the states after this to ensure the gripper is in the correct state for their behaviour.

robot_skills/src/robot_skills/arms.py
PetervDooren: Real life discussion with Matthijs: occupied by = None should be included in open gripper then. 
Technically occupied by should not be in the robot object.

PR #1051, opened 2020-10-20
Feature/arm requirements2

MatthijsBurgh:


robot_skills/src/robot_skills/robot_part.py
```diff
@@ -32,6 +33,15 @@ def __init__(self, robot_name, tf_listener):
         # If no such check exists, then assume it's operational unless overridden in subclass
         self._operational = True
 
+    @property
+    def id(self):
```
LoyVanBeek, line 37: This is something to look out with: `id` is also a Python thing itself, that is why it's called `_id` in the constructor. Any Python object already has an `id` but that is a built-in thing we do now want to mess with. An option is to name the method/property `id_`

MatthijsBurgh:
Tests shouldn't fix the order of the arm requirments.

robot_skills/src/robot_skills/arms.py
```diff
@@ -1027,3 +1028,60 @@ def object_in_gripper_measurement(self):
...
+        # Check if the child_state is a state_machine (must be done before checking for arm properties!)
+        if isinstance(child_state, smach.StateMachine):
+            child_sm_arm_requirements = collect_arm_requirements(child_state)
+            update_requirements(child_sm_arm_requirements)
+
+        # If no arm properties defined: continue
+        if not hasattr(child_state, "REQUIRED_ARM_PROPERTIES"):
+            continue
+
+        update_requirements(child_state.REQUIRED_ARM_PROPERTIES)
```
MatthijsBurgh, line 1067:
```suggestion
        if hasattr(child_state, "REQUIRED_ARM_PROPERTIES"):
            update_requirements(child_state.REQUIRED_ARM_PROPERTIES)
```

robot_skills/src/robot_skills/arms.py
```diff
@@ -4,6 +4,7 @@
 import rospy
 import std_msgs.msg
 import PyKDL as kdl
+import smach
```
MatthijsBurgh, line 7: It still feels weird to have `smach` in `robot_skills`.

If it stays here, smach should be added as dependency in `package.xml` and `CMakeLists.txt`

robot_skills/test/test_collect_arm_requirements.py
```diff
@@ -35,7 +36,8 @@ def __init__(self, robot):
             **self.REQUIRED_ARM_PROPERTIES), "None of the available arms meets all this class's" \
                                              "requirements: {}".format(self.REQUIRED_ARM_PROPERTIES)
 
-    def execute(self, userdata):
+    @staticmethod
+    def execute():
```
MatthijsBurgh, line 40: Dit mag niet, je moet wel `userdata` meegeven volgens.

robot_skills/test/test_collect_arm_requirements.py
```diff
@@ -0,0 +1,118 @@
...
+    @classmethod
+    def setUpClass(cls):
+        cls.robot = get_robot('mockbot')
+
+    def test_collect_arm_requirements(self):
+        requirements = collect_arm_requirements(FirstStateMachine(self.robot))
+        reference = {'required_trajectories': ['prepare_grasp', 'prepare_place'],
+                     'required_gripper_types': ['pseudo-gripper-type-any-grasping-will-do',
+                                                'gripper-type-pinch']}
+        compare_lists_in_dicts(reference, requirements)
```
PetervDooren, line 103: If I understand correctly then this function deletes the matching pairs between reference and requirements. So if everything works properly, both should be empty at the end correct?

Why did you make this test like this, rather than assert that the two are equal?

robot_smach_states/examples/manipulation/example_grasping.py
```diff
@@ -41,9 +41,7 @@
 
     item = ds.Designator(item)
 
-    arm = ds.UnoccupiedArmDesignator(robot, arm_properties={"required_trajectories": ["prepare_grasp"],
-                                                            "required_goals": ["carrying_pose"],
-                                                            "required_gripper_types": [arms.GripperTypes.GRASPING]})
+    arm = ds.UnoccupiedArmDesignator(robot, arm_properties={})
```
PetervDooren, line 44: do we still need to specify arm properties if it is empty

PetervDooren (changes requested):

robot_skills/test/test_collect_arm_requirements.py
LarsJanssenTUe: Yes that is completely true! I started out with an assert in which the two should be equal, but then @MatthijsBurgh pointed out that this would break in the case that (for some reason) the order of the requirements changes. So to allow for this I created this test. I think it is ugly, but I couldn't think of anything better. If you have any ideas let me know!

PetervDooren: One possibility is as follows: for every entry in results, check that it exists in reference. And following that for every entry in reference check that it exists in results. You may also look into comparing dicts in python. 
Make sure that in the error message you specify: complete results and reference and the difference which caused the error.

robot_skills/src/robot_skills/arms.py
LarsJanssenTUe: The easy way around this would be change the position of the collect_arm_requirements def. @PetervDooren also had a remark about this so I will see about moving it somewhere else.

robot_skills/test/test_collect_arm_requirements.py
LarsJanssenTUe: Wat bedoel je met "mag niet"? De test succeed namelijk wel, maar als het bad practice is dan zal ik userdata toevoegen.

robot_skills/src/robot_skills/arms.py
```diff
@@ -1027,3 +1028,60 @@ def object_in_gripper_measurement(self):
...
+
+    # Check arm requirements
+    arm_requirements = {}
+
+    for child_state in state_machine.get_children().itervalues():
+
+        # Check if the child_state is a state_machine (must be done before checking for arm properties!)
+        if isinstance(child_state, smach.StateMachine):
+            child_sm_arm_requirements = collect_arm_requirements(child_state)
+            update_requirements(child_sm_arm_requirements)
```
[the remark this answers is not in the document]
LarsJanssenTUe: Yes because a state in a statemachine can again be a statemachine

robot_skills/test/test_collect_arm_requirements.py
LarsJanssenTUe: That could indeed be an option, but why would you choose that over this? TBH I think this is clearer and faster, but maybe not cleaner..... @PetervDooren

MatthijsBurgh:
@LarsJanssenTUe The reason `robot_smach_states` can't be found in the `robot_skills` test, is that `robot_skills` doesn't depend on `robot_smach_states`. And this is also not possible, as a reverse dependency already exist. So the test regarding the collection of the arm requirements needs to be in `robot_smach_states` if both `robot_skills` and `robot_smach_states` components are needed.

robot_smach_states/src/robot_smach_states/navigation/navigation.py
```diff
@@ -40,14 +40,16 @@ def execute(self, userdata=None):
 
 
 class getPlan(smach.State):
-    def __init__(self, robot, constraint_function, speak=True):
+    def __init__(self, robot, constraint_function, input_keys=[], output_keys=[], speak=True):
```
MatthijsBurgh, line 43: Lists as default args are not allowed, see https://docs.python-guide.org/writing/gotchas/

robot_smach_states/src/robot_smach_states/navigation/navigation.py
```diff
@@ -216,7 +218,8 @@ class NavigateTo(smach.StateMachine):
     :param speak: Whether or not the robot should speak during navigation
     :param reset_pose: Whether or not the robot is allowed to change its pose for navigation.
     """
-    def __init__(self, robot, constraint_function, reset_head=True, speak=True, reset_pose=True, input_keys=[], output_keys=[]):
+    def __init__(self, robot, constraint_function, reset_head=True, speak=True, reset_pose=True, input_keys=[],
```
MatthijsBurgh, line 221: Lists as default argument, again

robot_smach_states/src/robot_smach_states/util/designators/arm.py
```diff
@@ -95,7 +78,7 @@ class UnoccupiedArmDesignator(ArmDesignator):
     >>> arm_to_use_for_third_grab = empty_arm_designator.resolve()
     >>> assert arm_to_use_for_third_grab is None
     """
-    def __init__(self, robot, arm_properties, name=None):
+    def __init__(self, robot, arm_properties={}, name=None):
```
MatthijsBurgh, line 81: Dict also can't be a default argument

robot_smach_states/src/robot_smach_states/utility.py
```diff
@@ -351,6 +353,109 @@ def execute(self, userdata=None):
         return "done"
 
 
+class ResolveArm(smach.State):
```
MatthijsBurgh, line 356: I really don't get why we have a state to resolve the arm after which we then use userdata to pass on the arm.

So why do we need to have a state to resolve the arm? Why can't this happen in a designator which is resolved by the first state that uses it?

Why do we use usedata after resolving the arm instead of a designator?

robot_smach_states/src/robot_smach_states/util/designators/arm.py
```diff
@@ -145,8 +128,8 @@ def _resolve(self):
         if not entity:
             rospy.logdebug('ArmHoldingEntityDesignator: Entity to find in the arm does not exist')
```
MatthijsBurgh, line 129: Why debug? This is a significant issue, this should be an error.

MatthijsBurgh (commented):
I really don't get why we have a state to resolve the arm after which we then use userdata to pass on the arm.

So why do we need to have a state to resolve the arm? Why can't this happen in a designator which is resolved by the first state that uses it?

Why do we use usedata after resolving the arm instead of a designator? IMO it should almost be impossible to use designators for this to introduce userdata in our design

robot_smach_states/src/robot_smach_states/utility.py
PetervDooren: With the smach state system we do need a state before navigate_to_grasp to put the arm into the userdata. That is a requirement.
As for userdata over designator. I remember some discussion between people smarter than me about this

robot_smach_states/src/robot_smach_states/utility.py
```diff
@@ -351,6 +353,109 @@ def execute(self, userdata=None):
...
+        :param state: Smach state of which arm requirements should be checked against the current arm requirements
+        :return: current arm requirements, updated (if necessary) given the input state
+        """
+        for k, v in state.iteritems():
+            if k not in arm_requirements:
+                arm_requirements[k] = v
+            else:
+                for value in v:
+                    if value not in arm_requirements[k]:
+                        arm_requirements[k] += v
```
PetervDooren, line 416: This is what I meant when I proposed to split some of the functionality of const_resolve. Because this loop is very similar to the one below. 
Having said that. This is a readable code nitpick and not a critisism of your arms requirements approach.

robot_smach_states/src/robot_smach_states/utility.py
```diff
@@ -351,6 +353,109 @@ def execute(self, userdata=None):
...
+    """
+
+    arm_designator_add_props = copy.copy(arm_designator)
+    for k, v in additional_properties.items():
+        if k in arm_designator_add_props.arm_properties:
+            for val in additional_properties[k]:
+                if val not in additional_properties[k]:
+                    arm_designator_add_props.arm_properties[k] += val
+        else:
+            arm_designator_add_props.arm_properties[k] = v
```
PetervDooren, line 455: This is the loop that is very similar and which might be best served as its own function.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -344,18 +345,24 @@ def __init__(self, robot, item, arm):
...
-            smach.StateMachine.add('GRAB', PickUp(robot, arm, item),
+            smach.StateMachine.add('GRAB', PickUp(robot, item),
                                    transitions={'succeeded': 'done',
                                                 'failed': 'RESET_FAILURE'})
 
-            smach.StateMachine.add("RESET_FAILURE", ResetOnFailure(robot, arm),
+            smach.StateMachine.add("RESET_FAILURE", ResetOnFailure(robot),
                                    transitions={'done': 'failed'})
+
+        check_arm_requirements(self, robot)
```
PetervDooren, line 368: Why check the arms requirement here rather than checking it at the very end of construction of your statemachine. just prior to execution.

robot_smach_states/src/robot_smach_states/utility.py
```diff
@@ -351,6 +353,109 @@ def execute(self, userdata=None):
...
+            else:
+                for value in v:
+                    if value not in arm_requirements[k]:
+                        arm_requirements[k] += v
+
+    # Check arm requirements
+    arm_requirements = {}
+
+    if isinstance(state_machine, smach.State):
+        if hasattr(state_machine, "REQUIRED_ARM_PROPERTIES"):
```
PetervDooren, line 422: One GIANT asterisk of this approach is that you assume that a state or smach state will only use one arm. And that there are no states which use two arms, each of which have different requirements.

PetervDooren (changes requested):

robot_skills/src/robot_skills/arm/grasp_sensor.py
```diff
@@ -0,0 +1,127 @@
+from robot_skills.robot_part import RobotPart
+from tue_manipulation_msgs.msg import GripperCommandGoal, GripperCommandAction
+from tue_msgs.msg import GripperCommand
+
```
jlunenburg, line 4: Insufficient whitespace

robot_skills/src/robot_skills/arm/grasp_sensor.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        Constructor
+
+        :param distance: float with measured distance
+        """
+        self._distance = distance
+        self._stamp = rospy.Time.now()
+
+        # If the grasp sensor distance is smaller than this value, the gripper is holding an object
+        self.GRASP_SENSOR_THRESHOLD = rospy.get_param("skills/arm/grasp_sensor/threshold", 0.1)
```
jlunenburg, line 23: Why are this all caps? Doesn't look nice

robot_skills/src/robot_skills/arm/grasp_sensor.py
```diff
@@ -0,0 +1,127 @@
...
+        Returns if the gripper is empty, i.e., the measurement is recent and the value is greater than the threshold
+
+        :return: bool if holding
+        """
+        return self._is_recent() and self._distance > self.GRASP_SENSOR_THRESHOLD
+
+    def __repr__(self):
+        return "Distance: {}, is_holding: {}, is_unknown: {}, " \
+               "is_empty: {}".format(self.distance, self.is_holding, self.is_unknown, self.is_empty)
+
```
jlunenburg, line 83: Whitespace

robot_skills/src/robot_skills/arm/grasp_sensor.py
LarsJanssenTUe: These commits are inherited from master, but I fixed the white spaces.

robot_smach_states/src/robot_smach_states/utility.py
LarsJanssenTUe: Yes that is true, but I guess this is an issue for when we have a robot with 2 arms ;). By the way then "simply" a new implementation can be made using the current setup.

robot_smach_states/src/robot_smach_states/utility.py
LarsJanssenTUe: From what I see these two pieces of code are not mergeable.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
LarsJanssenTUe: After discussion with @MatthijsBurgh we decided to leave it here.

PR #1055, opened 2020-12-01
Suction cup interface

PetervDooren:
I wanted to pick up cards with the robot. So I created an interface and a smach state to use it. More (detection) capabilities are required to actually pick up cards

### DIFF
--- a/robot_skills/package.xml
+++ b/robot_skills/package.xml
@@ -37,6 +37,7 @@
   <depend>tue_manipulation_msgs</depend>
   <depend>tue_msgs</depend>
   <depend>visualization_msgs</depend>
+  <depend>tmc_suction</depend>
 
   <exec_depend>python-mock</exec_depend>
 
--- a/robot_skills/src/robot_skills/hero.py
+++ b/robot_skills/src/robot_skills/hero.py
@@ -22,7 +22,7 @@ class Hero(robot.Robot):
 
         self.add_arm_part(
             'leftArm',
-            arms.SuctionGripperArm(self.robot_name, self.tf_listener, self.get_joint_states, "left")
+            arms.ForceSensingArm(self.robot_name, self.tf_listener, self.get_joint_states, "left")
         )
 
         self.add_body_part('head', head.Head(self.robot_name, self.tf_listener))
--- a/robot_skills/src/robot_skills/hero.py
+++ b/robot_skills/src/robot_skills/hero.py
@@ -22,7 +22,7 @@ class Hero(robot.Robot):
 
         self.add_arm_part(
             'leftArm',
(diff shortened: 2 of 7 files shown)


robot_skills/src/robot_skills/hero.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, wait_services=False):
 
         self.add_arm_part(
             'leftArm',
-            arms.ForceSensingArm(self.robot_name, self.tf_listener, self.get_joint_states, "left")
+            arms.SuctionGripperArm(self.robot_name, self.tf_listener, self.get_joint_states, "left")
```
PetervDooren, line 25: Currently an arm can only be a forcesensingArm or a SuctionGripperArm. This is a problem but not the problem for this PR. Until then either solution misses out on functionality. Luckily neither are really used in the common smach states

robot_skills/package.xml
```diff
@@ -37,6 +37,7 @@
   <depend>tue_manipulation_msgs</depend>
   <depend>tue_msgs</depend>
   <depend>visualization_msgs</depend>
+  <depend>tmc_suction</depend>
```
PetervDooren, line 40: It kinda sucks that I need to include a toyota specific package here. Because a robot that doesn't use suction cup arms doesnt need this dependency.

MatthijsBurgh: Indeed, this should be handled differntly.

PetervDooren: But how? The suction cup interface needs this message type. The only way I see this happening is if we are able to separate  each robot part in a separate repo so that the robot can be composed only of the parts that it has and use the dependencies of its parts only.

LoyVanBeek: I also don't like having this dependency, but I don't see any other way either. One way or another, we'll need those messages.

An option could be to have a separate package with a more general (non-TMC) suction cup interface, but for TMC that would still have to depend on `tmc_suction`. That conversion could be done in a bridge node, specific for HERO. That is a bit more complex due to the indirection.

robot_skills/src/robot_skills/hero.py
PetervDooren, line 25:
```suggestion
            arms.ForceSensingArm(self.robot_name, self.tf_listener, self.get_joint_states, "left")
```

PetervDooren: Reverted this change for now. But a change is needed to make the robot arm part composable of different functionality

MatthijsBurgh:
@PetervDooren is this still current?

PetervDooren:
This was left open until composable robot interface was fixed. Now this needs to be changed to fit that, which should make this a lot more elegant

MatthijsBurgh:
Ok, removed myself as reviewer. Request a review when ready.

PR #1060, opened 2020-12-21
(robot_skills) remove dep on amigo_msgs

MatthijsBurgh:


LarsJanssenTUe (commented):
By doing this we completely get rid of the compatability with amigo right? Is that what we want? @MatthijsBurgh @PetervDooren

MatthijsBurgh:
No that is not what is happening. This is an example. so not run in any test.

It is just that we remove the example and the dependency it creates

PetervDooren (approved):

PetervDooren:
make sure the tests work and this can be merged

PR #1058, opened 2020-12-12
Composable robot_interface

PetervDooren:

PR #1061, opened 2020-12-29
Update robot interface

PetervDooren:
Update robot smach states and challenges to conform to #1058 
Merge after #1058

### DIFF
--- a/robot_skills/src/robot_skills/mockbot.py
+++ b/robot_skills/src/robot_skills/mockbot.py
@@ -47,14 +47,13 @@ def old_query(spec, choices, timeout=10):
 
 
 class AlteredMagicMock(mock.MagicMock):
-    def assert_called_with(_mock_self, *args, **kwargs):
+    def assert_called_with(self, *args, **kwargs):
         """
         Assert that the mock was called with the specified arguments.
 
         Raises an AssertionError if the args and keyword args passed in are
         different to the last call to the mock.
         """
-        self = _mock_self
         if self.call_args is None:
             expected = self._format_mock_call_signature(args, kwargs)
             raise AssertionError('Expected call: %s\nNot called' % (expected,))
--- a/robot_skills/src/robot_skills/mockbot.py
+++ b/robot_skills/src/robot_skills/mockbot.py
@@ -84,9 +84,7 @@ class AlteredMagicMock(mock.MagicMock):
         if not any([self.equal_or_more_arguments(expected, c) for c in actual]):
             cause = expected if isinstance(expected, Exception) else None
             expected_string = self._format_mock_call_signature(args, kwargs)
-            six.raise_from(AssertionError(
-                '%s call not found' % expected_string
-            ), cause)
+            six.raise_from(AssertionError("'%s' call not found" % expected_string), cause)
 
     @staticmethod
     def equal_or_more_arguments(expected, actual):
--- a/robot_skills/src/robot_skills/mockbot.py
+++ b/robot_skills/src/robot_skills/mockbot.py
@@ -108,17 +108,17 @@ class MockedRobotPart(object):
         self.robot_name = robot_name
         self.tf_listener = tf_listener
 
-        self.load_param = mock.MagicMock()
-        self.wait_for_connections = mock.MagicMock()
-        self.create_simple_action_client = mock.MagicMock()
-        self.create_service_client = mock.MagicMock()
-        self.create_subscriber = mock.MagicMock()
-        self._add_connection = mock.MagicMock()
-        self.operational = mock.MagicMock()
-        self.subscribe_hardware_status = mock.MagicMock()
-        self.unsubscribe_hardware_status = mock.MagicMock()
-        self.process_hardware_status = mock.MagicMock()
-        self.reset = mock.MagicMock()
+        self.load_param = AlteredMagicMock()
+        self.wait_for_connections = AlteredMagicMock()
+        self.create_simple_action_client = AlteredMagicMock()
+        self.create_service_client = AlteredMagicMock()
+        self.create_subscriber = AlteredMagicMock()
+        self._add_connection = AlteredMagicMock()
+        self.operational = AlteredMagicMock()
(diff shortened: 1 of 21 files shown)


robot_smach_states/test/manipulation/test_manipulation.py
```diff
@@ -11,6 +11,17 @@
...
+def assert_any_first_argument(unit_test_case, calls, argument):
+    for call in calls:
+        if call[0][0] == argument:
+            return
+    raise AssertionError("co call in {} has first argument {}".format(calls, argument))
+
+
+def assert_first_argument(unit_test_case, call, argument):
+    unit_test_case.assertEqual(call[0][0], argument)
+
```
PetervDooren, line 23: You may wonder why these custom functions are necessary. That is because the existing tests were too strict. A call with arguments ('open') would not pass because we require additional arguments in this function. In practice these additional arguments are provided by default arguments.

robot_smach_states/test/manipulation/test_manipulation.py
```diff
@@ -11,6 +11,17 @@
 from robot_smach_states.util import designators as ds
 
 
+def assert_any_first_argument(unit_test_case, calls, argument):
+    for call in calls:
+        if call[0][0] == argument:
+            return
+    raise AssertionError("co call in {} has first argument {}".format(calls, argument))
```
MatthijsBurgh, line 18: `co call`?

robot_smach_states/test/manipulation/test_manipulation.py
```diff
@@ -11,6 +11,17 @@
 from robot_smach_states.util import designators as ds
 
 
+def assert_any_first_argument(unit_test_case, calls, argument):
+    for call in calls:
+        if call[0][0] == argument:
```
MatthijsBurgh, line 16: Are you sure, that `call[0][0]` is always available? Aka do the list always have at least one element in them?

robot_smach_states/test/manipulation/test_manipulation.py
MatthijsBurgh: I think these functions are related to our mockbot, so might be better to move these to there.

MatthijsBurgh (changes requested):

robot_smach_states/test/manipulation/test_manipulation.py
PetervDooren:
```suggestion
    raise AssertionError("no call in {} has first argument {}".format(calls, argument))
```

robot_smach_states/test/manipulation/test_manipulation.py
PetervDooren: No, if the call has no arguments then call[0] is an empty list. But since this is a test it is allowed to fail. the test will still give an accurate result

robot_smach_states/test/manipulation/test_manipulation.py
PetervDooren: They are related to mock in general. So yes I can definitly move these to the mockbot.

robot_smach_states/test/manipulation/test_manipulation.py
MatthijsBurgh: Might be helpful to include this in option in the function to raise a more accurate error?

PetervDooren: Will do

robot_skills/src/robot_skills/mockbot.py
```diff
@@ -45,22 +46,81 @@ def old_query(spec, choices, timeout=10):
     raise Exception('robot.ears.recognize IS REMOVED. Use `robot.hmi.query`')
 
 
+class AlteredMagicMock(mock.MagicMock):
```
PetervDooren, line 49: I created an Altered magicMock class to allow us to create less strict tests. This way a call:
mock.('argument 1', 'argument_2', kwarg1='keywordargument 1') will satisfy
mock.assert_called_with('argument 1'). This way we can relax requirements on the amount of arguments passed. So you can rely on the default arguments or you can add your own. This mock allows both

robot_smach_states/test/manipulation/test_manipulation.py
PetervDooren: I may have gone overboard... 
see [14bc248](https://github.com/tue-robotics/tue_robocup/pull/1061/commits/14bc2486e91cc52f81d47c67b03fd333e4682d10)

robot_skills/src/robot_skills/mockbot.py
MatthijsBurgh: Maybe we could monkey patch MagicMock

robot_skills/src/robot_skills/mockbot.py
```diff
@@ -7,6 +7,7 @@
 import mock
 import random
 import os
+import six
```
MatthijsBurgh, line 10: This should be added to package.xml

MatthijsBurgh (changes requested):

robot_skills/src/robot_skills/mockbot.py
PetervDooren: I don't think monkeypatching would be elegant in this scenario

robot_skills/src/robot_skills/mockbot.py
PetervDooren: done

robot_skills/src/robot_skills/util/kdl_conversions.py
```diff
@@ -87,6 +87,9 @@ def __eq__(self, other):
         else:
             return False
 
+    def __ne__(self, other):
+        return not self.__eq__(other)
```
PetervDooren, line 91: CI wasn't working. I found out __ne__ is not defined for the vector stamped class. Could place this in another PR but its a small fix

MatthijsBurgh (approved):
I don't like the old style string substitution for the rest this is good to go.

Please check my changes.

PR #1070, opened 2021-02-06
added battery object

PetervDooren:


MatthijsBurgh (changes requested):
Please add some docstrings.

This only works for HERO. As the battery topics on the TUe robots are different. That should also be documented.

Also this implementation assumes that the last message is indeed recent. If the publisher stops publishing, the last message gets outdated. I am fine with the implementation, but it should documented.

Final question: What is going to be used for?

PetervDooren:
Preventing the robot from driving off while it is charging:)

robot_skills/src/robot_skills/battery.py
```diff
@@ -0,0 +1,28 @@
...
+
+from sensor_msgs.msg import BatteryState
+from robot_skills.robot_part import RobotPart
+
+
+class Battery(RobotPart):
+    def __init__(self, robot_name, tf_listener, location):
+        super(Battery, self).__init__(robot_name=robot_name, tf_listener=tf_listener)
+        self.location = location
+        self.battery_sub = rospy.Subscriber("battery", BatteryState, self.battery_callback)
```
MatthijsBurgh, line 19: Please use https://github.com/tue-robotics/tue_robocup/blob/master/robot_skills/src/robot_skills/robot_part.py#L130

MatthijsBurgh (changes requested):

MatthijsBurgh (approved):

Issue #1071, opened 2021-02-06
purge leftArm


PetervDooren:
https://github.com/tue-robotics/tue_robocup/blob/ecf951a46af3a1205bcacd9fffd0d2652d2bafef/robot_skills/src/robot_skills/robot.py#L131

the 'leftArm' and 'rightArm' properties of the robot object should be removed. There are a few placed in the robot_skills and robot_smach states where these are still used however. This should be fixed.

PetervDooren:
solved with #1073

PR #1073, opened 2021-02-06
Fix/hardcoded armparts

PetervDooren:


PetervDooren:
solves #1071

PetervDooren:
This issue was not solved in challenge dishwasher and other challenges which use this.

LarsJanssenTUe (approved):

PR #1083, opened 2021-02-19
Purge 'side' parameter

PetervDooren:
We dont want to refer to left or right arms because not every robot has two arms which are symetrical. For that reason we want to remove the parameter 'side' and replace it with something more general

Must be merged together with: https://github.com/tue-robotics/hero_description/pull/11 and https://github.com/tue-robotics/hero_bridge/pull/23

I think this should be tested on the real robot, just to be safe...


robot_skills/src/robot_skills/arm/arms.py
```diff
@@ -265,67 +265,64 @@ def _test_die(self, cond, feature, hint=''):
...
-
-    Examples:
-    >>> left.send_goal(0.265, 1, 0.816, 0, 0, 0, 60)  # doctest: +SKIP
-    or Equivalently:
-    >>> left.send_goal(px=0.265, py=1, pz=0.816, yaw=0, pitch=0, roll=0, time_out=60, pre_grasp=False, frame_id='/amigo/base_link')  # doctest: +SKIP
+    A kinematic chain ending in an end_effector. Can be controlled using either joint goals or a goal to reach with
+    the end_effector described in carthesian coordinates.
     """
-    def __init__(self, robot_name, tf_listener, get_joint_states, side):
+    def __init__(self, robot_name, tf_listener, get_joint_states, name):
```
PetervDooren, line 279: To make this interface truly generic we should make it so that not all parameters, actions, tf_frames are linked via the arm name. Then we should allow these things as parameters to the interface. But that might be something for another day...

MatthijsBurgh (dismissed):

robot_skills/package.xml
```diff
@@ -41,6 +41,7 @@
 
   <test_depend>test_tools</test_depend>
   <test_depend>python-six</test_depend>
+  <test_depend>hero_description</test_depend>
```
MatthijsBurgh, line 44: I think this is a very dirty dependency. But this might be fixed by @jlunenburg.

robot_skills/src/robot_skills/arm/arms.py
```diff
@@ -265,67 +265,64 @@ def _test_die(self, cond, feature, hint=''):
...
 
         :param robot_name: robot_name
         :param tf_listener: tf_server.TFClient()
         :param get_joint_states: get_joint_states function for getting the last joint states
-        :param side: left or right
+        :param name: string used to identify the arm
         """
         super(Arm, self).__init__(robot_name=robot_name, tf_listener=tf_listener)
-        self.side = side
+        self.arm_name = name
```
MatthijsBurgh, line 289: Why have `name` as argument, but store it as `arm_name`. I think it is better to use `name` inside the class.

robot_skills/src/robot_skills/arm/gripper.py
```diff
@@ -68,7 +68,7 @@ def __init__(self, robot_name, tf_listener, gripper_name):
         """
         super(ParrallelGripper, self).__init__(robot_name=robot_name, tf_listener=tf_listener)
         self.gripper_name = gripper_name
-        offset = self.load_param('skills/arm/' + self.gripper_name + '/grasp_offset/')
+        offset = self.load_param('skills/' + self.gripper_name + '/grasp_offset/')
```
MatthijsBurgh, line 71: I would say a gripper is part of an arm, so the parameter is also part in the parameter tree of the arm.

robot_skills/test/rostest/test_hero.py
```diff
@@ -0,0 +1,23 @@
...
+
+# robot skills
+from robot_skills.hero import Hero
+
+
+class TestHeroInterface(unittest.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
+        rospy.init_node('test_hero')
```
MatthijsBurgh, line 14: I think it is common to do this in the `__main__`

MatthijsBurgh (changes requested):

robot_skills/package.xml
PetervDooren: It is dirty. But it is a dependency that we have always had implicitly as we cant construct a robot object without its description.

robot_skills/src/robot_skills/arm/arms.py
PetervDooren: Agreed

robot_skills/src/robot_skills/arm/gripper.py
PetervDooren: It is not necessarily part of an arm. That is why we created a more composable robot interface

robot_skills/src/robot_skills/arm/arms.py
```diff
@@ -265,67 +265,64 @@ def _test_die(self, cond, feature, hint=''):
...
-        self.side = side
+        self.name = name
 
         self._operational = True  # In simulation, there will be no hardware cb
 
         # Get stuff from the parameter server
-        offset = self.load_param('skills/arm/' + self.side + '/grasp_offset/')
+        offset = self.load_param('skills/gripper/grasp_offset/')
         self.offset = kdl.Frame(kdl.Rotation.RPY(offset["roll"], offset["pitch"], offset["yaw"]),
                                 kdl.Vector(offset["x"], offset["y"], offset["z"]))
```
MatthijsBurgh, line 296: Shouldn't this be gripper specific

robot_skills/src/robot_skills/arm/grasp_sensor.py
```diff
@@ -87,20 +91,18 @@ class GraspSensor(RobotPart):
...
         :param tf_listener: tf_server.TFClient()
-        :param side: left or right arm of the robot.
+        :param sensor_topic: name of the topic where measurements are published
         """
         super(GraspSensor, self).__init__(robot_name=robot_name, tf_listener=tf_listener)
         # Init grasp sensor subscriber
         self._grasp_sensor_state = GripperMeasurement(0.0)
-        rospy.Subscriber("/" + robot_name + "/" + side + "_arm/proximity_sensor",
-                         std_msgs.msg.Float32MultiArray, self._grasp_sensor_callback)
+        rospy.Subscriber("/" + robot_name + "/" + sensor_topic, Float32MultiArray, self._grasp_sensor_callback)
```
MatthijsBurgh, line 105: Now we force the robot name to be in the topic. Probably that is always the case, but I prefer we don't apply any unesesarry contraints on the topic. Let the user define the entire topic.

MatthijsBurgh (changes requested):

robot_skills/src/robot_skills/arm/arms.py
PetervDooren: It should be, but it is outside the scope of this PR

robot_skills/src/robot_skills/arm/grasp_sensor.py
PetervDooren: Almost all topics within the interface are prefixed with the robots name. I think it would cause more trouble than its worth at this stage. There are limits to how generic we want to make these interfaces. And being able to quickly change the prefix is an important part of making an interface for a new robot.

robot_skills/src/robot_skills/arm/arms.py
```diff
@@ -265,67 +265,64 @@ def _test_die(self, cond, feature, hint=''):
...
+            "/" + robot_name + "/" + self.name + "/grasp_precompute", GraspPrecomputeAction)
 
         # Init joint trajectory action server
         self._ac_joint_traj = self.create_simple_action_client(
             "/" + robot_name + "/body/joint_trajectory_action", FollowJointTrajectoryAction)
 
         # Init marker publisher
         self._marker_publisher = rospy.Publisher(
-            "/" + robot_name + "/" + self.side + "_arm/grasp_target",
+            "/" + robot_name + "/" + self.name + "/grasp_target",  # TODO: update rviz config
```
MatthijsBurgh, line 325: This shouldn't be in the code.

robot_skills/src/robot_skills/arm/grasp_sensor.py
MatthijsBurgh: But I am pretty sure that other robot parts don't do this. So please keep it consistent.

MatthijsBurgh (changes requested):

robot_skills/src/robot_skills/arm/arms.py
PetervDooren: Yes will update the rviz config and remove this

MatthijsBurgh: I suggest to edit the config in an editor instead of rviz. This prevents any changes somewhere else in the config.

robot_skills/src/robot_skills/arm/grasp_sensor.py
MatthijsBurgh: @PetervDooren I don't think we do this somewhere else. So therefore I still think we should change this.

MatthijsBurgh (changes requested):
Only one remaining conversation

PetervDooren:
I don't think the the fact that we dont do this like that anywhere else is reason to change this. 
I assume by 'this' you mean having a sensor topic as input?

MatthijsBurgh:
I don't mind having the sensor topic as argument. I only mind that we force a prefix of the robot name on the topic.

PetervDooren:
That is something we do with every topic/action/service. That is not something unusual in robot_skills

MatthijsBurgh:
No, we don't do this in `robot_skills`. Not for the ones that have the topic as the argument. I.E. https://github.com/tue-robotics/tue_robocup/blob/9bece905c6b661a9c4f95d8d6dde2c6b5c203531/robot_skills/src/robot_skills/perception.py#L23-L31

MatthijsBurgh (changes requested):
Please rebase before approval

robot_skills/package.xml
```diff
@@ -47,6 +47,7 @@
   <exec_depend condition="$ROS_PYTHON_VERSION == 3">python3-mock</exec_depend>
 
   <test_depend>test_tools</test_depend>
+  <test_depend>hero_description</test_depend>
```
MatthijsBurgh, line 50: I think this one can be removed now with the seperated skills

MatthijsBurgh (changes requested):

robot_skills/package.xml
PetervDooren: sharp, I missed that one

PetervDooren:
```suggestion
```

MatthijsBurgh (approved):
Wait for CI before merging

robot_skills/src/robot_skills/torso.py
```diff
@@ -24,12 +24,16 @@ def __init__(self, robot_name, tf_listener, get_joint_states):
         super(Torso, self).__init__(robot_name=robot_name, tf_listener=tf_listener)
 
         self.joint_names = self.load_param('skills/torso/joint_names')
-        self._arm_joint_names = self.load_param('skills/arm/joint_names')
         self.default_configurations = self.load_param('skills/torso/default_configurations')
         self.default_tolerance = self.load_param('/skills/torso/default_tolerance')
         self.lower_limit = self.default_configurations['lower_limit']
         self.upper_limit = self.default_configurations['upper_limit']
 
+        try:  # TODO make sure this is no longer needed
```
jlunenburg, line 32: Why not load them outside and provide them to the torso via the constructor?

This solution isn't very nice because of two reasons:
* the try-except is error-prone. I'm not sure what'll happen if you move the torso
* it makes the torso depend on the fact that your arm is called 'arm_center', which is specific to Hero. So while making the Hero object more generic by removing 'left', you make this more specific by adding 'arm_center'

PetervDooren: agreed, but this is a hotfix for hero anyway. The only change made here is that the dependency on some arm is made explicit.
A different PR should make this parameter no longer necessary in the first place. So that we can send only one joint goal while keeping all others the same.

jlunenburg: Can I see that :blush: ? I like keeping the master branch as clean as possible and these hotfixes do not contribute to that.

PetervDooren: You are right. I will make this cleaner. Probably by allowing arm joint names to be passed in construction, rather than in the part itself

PetervDooren:
All tests are passed when all repositories are checked out to fix/side_param
see https://dev.azure.com/tue-robotics/tue_robocup/_build/results?buildId=2494&view=results

robot_skills/src/robot_skills/torso.py
```diff
@@ -14,7 +14,7 @@
 
 
 class Torso(RobotPart):
-    def __init__(self, robot_name, tf_listener, get_joint_states):
+    def __init__(self, robot_name, tf_listener, get_joint_states, arm_joint_names=[]):
```
MatthijsBurgh, line 17: Can't have a mutable default argument

MatthijsBurgh (changes requested):

robot_skills/src/robot_skills/torso.py
PetervDooren: fixed

MatthijsBurgh (approved):

PR #1088, opened 2021-02-23
Feature/use battery part

PetervDooren:
Created for people who occasionally forget that the robot is charging :)


PetervDooren:
I moved a lot of checks to a new state 'prepare drive' which ensures that the robot is ready to start driving. The one thing that could not be placed here is going to the driving pose, because that might change based on the distance we need to drive.

This does add the assumption that the things in the 'prepare drive' state only need to be checked at the start of navigation and cannot be changed during the process. i.e. noone plugs in the robot while navigating and noone uses and then releases the head.

robot_smach_states/src/robot_smach_states/navigation/navigation.py
```diff
@@ -39,6 +40,45 @@ def execute(self, userdata=None):
...
+
+        # Cancel head goal, we need it for navigation :)
+        if self.reset_head:
+            self.robot.head.close()
+
+        for part in self.robot.parts.itervalues():
+            # ensure all batteries are unplugged
+            if isinstance(part, Battery):
+                while part.is_charging:
+                    if tstart - rospy.Time.now() < rospy.Duration.from_sec(self.timeout):
```
MatthijsBurgh, line 70: I am not sure this needs to be and `rospy.duration`. Maybe a float/int would also work

MatthijsBurgh: First part is going to b e negative.... I don't think that should be the case. Also I prefer now > start+timeout

robot_smach_states/src/robot_smach_states/navigation/navigation.py
```diff
@@ -39,6 +40,45 @@ def execute(self, userdata=None):
...
+                        return 'abort'
+                    if self.preempt_requested():
+                        return 'preempted'
+                    rospy.logwarn("Robot is charging. Please unplug")
+                    self.robot.speech.speak("I am still charging. Please unplug me.")
+                    rospy.sleep(1)
+                rospy.logerr("Robot is charging, aborting navigation")
+                return 'abort'
```
MatthijsBurgh, line 78: This code is reaced when not charging anymore. Am I correct? That is not matching the code.

robot_smach_states/src/robot_smach_states/navigation/navigation.py
```diff
@@ -80,13 +120,12 @@ def execute(self, userdata=None):
 
 
 class executePlan(smach.State):
-    def __init__(self, robot, breakout_function, blocked_timeout=4, reset_head=True, reset_pose=True):
-        smach.State.__init__(self, outcomes=['succeeded', 'arrived', 'blocked', 'preempted'])
+    def __init__(self, robot, breakout_function, blocked_timeout = 4, reset_pose=True):
```
MatthijsBurgh, line 123: No space around equal sings in default arguments

MatthijsBurgh (changes requested):
Also fix CI

robot_smach_states/src/robot_smach_states/navigation/navigation.py
```diff
@@ -80,13 +118,12 @@ def execute(self, userdata=None):
 
 
 class executePlan(smach.State):
-    def __init__(self, robot, breakout_function, blocked_timeout=4, reset_head=True, reset_pose=True):
```
MatthijsBurgh, line 83: Why dropping this? Is this only obsolete for HERO or all robots?

MatthijsBurgh (changes requested):
CI fails on missing `reset_head`

MatthijsBurgh:
I think we can delete the branch. GH will still keep the diff.

PR #1094, opened 2021-04-01
noetic-devel

MatthijsBurgh:


jlunenburg (approved):
I don't think we should be nitpicky here about changes in commented stuff. I do consider that something we should keep a closer eye on in the future.

PR #1096, opened 2021-04-10
Register robot

### LINKED ISSUE #1095: Register robots to robot skills
Currently, the `get_robot` module in robot skills tries to import HERO, AMIGO and SERGIO. This solution is not scalable. To remove the (backwards) dependency (which is necessary to allow new deriving robots _without_ modifying `robot_skills`), some kind of registration method is required. 

For inspiration, we could look at [the class-registry package](https://pypi.org/project/class-registry/). However, we need to check:
1. How this works with the catkin `__init__` magic
2. How this works if we don't install packages (we usually run stuff directly from our source space)

Better still, [roswtf](https://wiki.ros.org/roswtf) also has a plugin system. Looking at [this implementation](https://github.com/ros/ros_comm/blob/noetic-devel/utilities/roswtf/src/roswtf/plugins.py) it shouldn't be difficult to reproduce this.

MatthijsBurgh:
Fixes #1095

### DIFF
--- a/robot_skills/README.md
+++ b/robot_skills/README.md
@@ -1,6 +1,17 @@
 # robot_skills
 
-Robot_skills provide interfaces to all parts of a robot: its base, arms, head, perception, worldmodel, speech system, etc. 
-The package also provides a console for a robot, so you can give the robot commands like goto(<some coordinate>) via a console
+`robot_skills` provide interfaces to all parts of a robot: its base, arms, head, perception, worldmodel, speech system, etc.
 
 A robot is represented via a robot-object which can be used by the [robot_smach_states](https://github.com/tue-robotics/tue_robocup/tree/master/robot_smach_states)
+
+This package contains a [`mockbot`](src/robot_skills/mockbot.py), which expose all the needed interfaces.
+
+## Robot registration
+
+To implement your own robot, create a robot specific skills package. To register your robot add the following to the `package.xml` of your package:
+```xml
+<export>
+  <robot_skills robot="bla_skills.bla.Bla" />
+</export>
+```
+The value of the robot attribute has to be of the format: `<package_name>.<sub_module>.<class_name>`.
--- a/robot_skills/src/robot_skills/get_robot.py
+++ b/robot_skills/src/robot_skills/get_robot.py
@@ -84,6 +84,6 @@ def get_robot(name):
     else:
         error_msg = "Cannot construct robot '{}'\n".format(name)
         error_msg += "Available robots:\n\t{}\n".format("\n\t".join(list(ROBOTS.keys())))
-        error_msg += "To install, try: 'tue-get install {}_skills'".format(name)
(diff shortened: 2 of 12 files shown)


PetervDooren (approved):

robot_skills/README.md
```diff
@@ -1,6 +1,17 @@
...
 A robot is represented via a robot-object which can be used by the [robot_smach_states](https://github.com/tue-robotics/tue_robocup/tree/master/robot_smach_states)
+
+This package contains a [`mockbot`](src/robot_skills/mockbot.py), which expose all the needed interfaces.
+
+## Robot registration
+
+To implement your own robot, create a robot specific skills package. To register your robot add the following to the `package.xml` of your package:
+```xml
+<export>
+  <robot_skills robot="bla_skills.bla.Bla" />
```
jlunenburg, line 14: More common would be:
```xml
<robot_skills robot="foo_skills.foo.Foo"/>
```

robot_skills/README.md
```diff
@@ -1,6 +1,17 @@
...
+</export>
+```
```
jlunenburg, line 16: It would also be valuable to refer to the 'unittest'. I'm planning on expanding this so that in principle if your robot passes the test, it can be used in smach states.

robot_skills/src/robot_skills/get_robot.py
```diff
@@ -1,32 +1,54 @@
...
+
+def load_robots():
+    """
+    Register the robots, which are export via package.xml 'export' tag
+    """
+    rospack = rospkg.RosPack()
+    to_check = rospack.get_depends_on('robot_skills', implicit=False)
+    to_check.append("robot_skills")  # also check for robots in robot_skills
+
+    for pkg in to_check:
```
jlunenburg, line 25: Not much to add here except some minor issues:
* you could slightly improve readability by putting part(s) of `load_robots` in different functions, e.g., `register_robot`
* thinking about that: what happens if you register two robots with the same name? You won't notice that the behavior is not what you expected (if you are the first user unaware of the second)

Styling
```python
rospy.logerr(
    "Unable to load robot '{}' from package '{}'. Exception thrown: '{}'".format(robot, pkg, e)
)
``` 
or
```python
rospy.logerr(
    "Unable to load robot '{}' from package '{}'. Exception thrown: '{}'".format(
        robot, pkg, e
    )
)
```

robot_skills/src/robot_skills/get_robot.py
```diff
@@ -1,32 +1,54 @@
...
+        m = rospack.get_manifest(pkg)
+        robots = m.get_export('robot_skills', 'robot')
```
LoyVanBeek, line 27: This is super nice!

robot_skills/README.md
LoyVanBeek, line 14: Nice work!

LoyVanBeek (commented):
This was a much needed refactoring, nice work folks :-D

PR #1097, opened 2021-04-11
use get_robot

PetervDooren:
As robot skills and hero/amigo skills have been moved the robot console no longer works. This change should fix it.

I should note that I was unable to test this yet, as hero_skills is somehow not recognised in my system.


MatthijsBurgh:
@PetervDooren I have also fixed this in #1096. Please review that one.

MatthijsBurgh:
This is is fixed by #1096

PR #1105, opened 2021-04-29
dont use tf prefix

PetervDooren:


MatthijsBurgh:
This is not really tf_prefix. On AMIGO/SERGIO, the links are just called ROBOT_NAME/LINK_NAME, so then the frames are also ROBOT_NAME/LINK_NAME. tf_prefix is that the links are called LINK_NAME, but the frames are called ROBOT_NAME/LINK_NAME.

So this should be configurable. This is also done, because the `base_link_frame` attribute of the robot is set per robot. Though this is not available in the robot parts...

PetervDooren:
I see, for HERO. the frames currently dont have the ROBOT_NAME/LINK_NAME format correct. This was a problem on the robot, where pickup failed since the 'hero/hand_palm_link' could not be found.

MatthijsBurgh:
Indeed on hero the links and thus the frames are just called LINK_NAME

robot_skills/src/robot_skills/arm/arms.py
```diff
@@ -490,8 +490,8 @@ def send_goal(self, frameStamped, timeout=30, pre_grasp=False, first_joint_pos_o
                 execute_timeout=rospy.Duration(timeout))
             if result == GoalStatus.SUCCEEDED:
 
-                result_pose = self.tf_listener.lookupTransform(self.robot_name + "/base_link",
-                                                               self.robot_name + "/" + self.grasp_frame,
+                result_pose = self.tf_listener.lookupTransform("base_link",
```
PetervDooren, line 493:
```suggestion
                result_pose = self.tf_listener.lookupTransform(self.robot_name + "/base_link",
```

MatthijsBurgh (approved):

PR #1123, opened 2021-05-19
added grab test

PetervDooren:


PetervDooren:
CI should fail now
1) pr #1121 has not yet been merged here
2) expecting problems because this added rostest falls within the scope of the nosetests.

robot_skills/src/robot_skills/mockbot.py
```diff
@@ -193,6 +193,7 @@ def __init__(self, robot_name, tf_buffer, *args, **kwargs):
         self.local_planner = AlteredMagicMock()
         self.local_planner.getStatus = mock.MagicMock(return_value="arrived")  # always arrive for now
         self.global_planner.getPlan = mock.MagicMock(return_value=["dummy_plan"])  # always arrive for now
+        self.global_planner.path_length = 0.0
```
PetervDooren, line 196: This is needed because navigation will fail otherwise. As it would try to compare a magicmock and a number

robot_smach_states/CMakeLists.txt
```diff
@@ -14,5 +14,7 @@ catkin_python_setup()
 catkin_package()
 
 if (CATKIN_ENABLE_TESTING)
+  find_package(rostest REQUIRED)
   catkin_add_nosetests(test)
+  add_rostest(test/rostest/test_grab.test)
```
PetervDooren, line 19: For some reason I did not need to separate rostests from nosetests like in robot_skills. I dont know why though

robot_skills/src/robot_skills/mockbot.py
MatthijsBurgh: Should this not also be a MagicMock?

MatthijsBurgh:
The mock path_length broke the tests of `robot_smach_states`...

robot_smach_states/CMakeLists.txt
MatthijsBurgh: Probably because there is no`__init__.py` in the `rostest` folder

MatthijsBurgh (approved):

PR #1130, opened 2021-06-22
Fix place

PetervDooren:


robot_smach_states/test/rostest/test_place.py
```diff
@@ -31,7 +31,9 @@ def test_place(self):
         itemdes = ds.Designator(item)
         arm = ds.ArmDesignator(self.robot).lockable()
 
-        state = Place(self.robot, itemdes, pose, arm)
+        posedes = ds.Designator(pose)
```
MatthijsBurgh, line 34: I think we should use `snake_case` for variable naming. I think that is much clearer than current naming convention.

MatthijsBurgh (changes requested):

MatthijsBurgh (approved):

PR #1142, opened 2021-10-10
allow direct access to arm parts

PetervDooren:
Previously we restricted access to the arm parts of the robot. This was so that challenge developers could not introduce robot-specific solutions. While I think this is a good philosophy for reusable solutions like the smach states I think for challenges we should make the direct robot interface more user friendly. 

We have often bypassed this restriction in challenges by simply calling the arm via hero.parts['center_arm']

### DIFF
--- a/robot_skills/src/robot_skills/robot.py
+++ b/robot_skills/src/robot_skills/robot.py
@@ -99,9 +99,8 @@ class Robot(object):
         :param arm_name: Name of the arm part.
         :param arm_part: Arm part object
         """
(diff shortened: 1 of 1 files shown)


PetervDooren:
That is only a hack when you want to make something generic. In practice we often want to make robot specific functionality in challenges.
And people who dont know this hack struggle to use one of the main features of the robot.
I dont think this will become a habit because the parts that we want to keep generic are tested with mockbot.

PR #1147, opened 2021-11-04
Apply ed_python

MatthijsBurgh:
Merge together with https://github.com/tue-robotics/action_server/pull/154


PetervDooren (changes requested):
Why the renaming of id to uuid. That will only cause confusion. I would keep it as is.

MatthijsBurgh:
Because this is also changed in `ed_python`. Because `id` (and  `type`) are core python functions. So we should not use these.

PetervDooren (approved):

PR #1160, opened 2022-01-04
add impuls environment

PetervDooren:
Files were just copied from robotics_testlab environment

### DIFF
--- a/robocup_knowledge/src/robocup_knowledge/environments/impuls/challenge_demo.py
+++ b/robocup_knowledge/src/robocup_knowledge/environments/impuls/challenge_demo.py
@@ -1,0 +1,189 @@
+from __future__ import print_function
+
+from robocup_knowledge import knowledge_loader
+common = knowledge_loader.load_knowledge("common")
+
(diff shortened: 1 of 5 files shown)


PetervDooren:
The whole robocup knowledge thing is a mess. It combines multiple things that have no business being together such as knowledge about an environement used for reasoning, speech recognition grammar, properties of object catagories and semantic properties of specific furniture.
We need to clean this up at some point because this is not maintainable.

MatthijsBurgh:
Maybe we should start by moving all the common functions to a shared file. Instead of copying them for each env.

PetervDooren:
That is not a step we can make now due to lack of resources (my time). 
I think we should start by thinking this over carefully.

LarsJanssenTUe:
I agree with Peter. I think a redesign can and should be done, before which I don't think it helps to start making these kind of changes. Therefore I will accept this PR, in order to unblock any potential development.

LarsJanssenTUe (approved):

PR #1163, opened 2022-02-19
Fixed example_hear_grap.py

shanki98:
Updated example to not use ears.recognize. If ears part is not used anymore, can we remove it from Master?

This example is additionally not complete because it does not search for the object.

### DIFF
--- a/robot_smach_states/examples/example_hear_grasp.py
+++ b/robot_smach_states/examples/example_hear_grasp.py
@@ -13,13 +13,13 @@ class Hear(smach.State):
         self.robot = robot
 
     def execute(self, userdata):
-        result = self.robot.ears.recognize(spec='Bring me a <drink>',
-                                           choices={'drink': ['coke', 'fanta']},
-                                           time_out=rospy.Duration(60))
-        choices = result.choices['drink'] if result else None
-        if choices:
-            rospy.loginfo('I heared:' + repr(choices))
-        userdata.heared = choices
+        answer = self.robot.hmi.query(description='can i have a coke?',
+                                      grammar="T -> DET coke \n DET -> a",
+                                      target='T',
+                                      timeout=rospy.Duration(60).to_sec())
+        if answer:
+            rospy.loginfo('I heard:' + answer.sentence)
+        userdata.heared = answer.sentence
         return 'succeeded'
 
 
--- a/robot_smach_states/examples/example_hear_grasp.py
+++ b/robot_smach_states/examples/example_hear_grasp.py
@@ -9,17 +9,17 @@ from robot_smach_states.utility import Initialize
 class Hear(smach.State):
     def __init__(self, robot):
         smach.State.__init__(self, outcomes=['succeeded', 'failed', 'preempted'],
-                             output_keys=['heared'])
+                             output_keys=['heard'])
         self.robot = robot
 
     def execute(self, userdata):
         answer = self.robot.hmi.query(description='can i have a coke?',
                                       grammar="T -> DET coke \n DET -> a",
                                       target='T',
-                                      timeout=rospy.Duration(60).to_sec())
+                                      timeout=60)
         if answer:
             rospy.loginfo('I heard:' + answer.sentence)
-        userdata.heared = answer.sentence
+        userdata.heard = answer.sentence
         return 'succeeded'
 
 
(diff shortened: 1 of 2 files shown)


robot_smach_states/examples/example_hear_grasp.py
```diff
@@ -13,13 +13,13 @@ def __init__(self, robot):
...
-                                           choices={'drink': ['coke', 'fanta']},
-                                           time_out=rospy.Duration(60))
-        choices = result.choices['drink'] if result else None
-        if choices:
-            rospy.loginfo('I heared:' + repr(choices))
-        userdata.heared = choices
+        answer = self.robot.hmi.query(description='can i have a coke?',
+                                      grammar="T -> DET coke \n DET -> a",
+                                      target='T',
+                                      timeout=rospy.Duration(60).to_sec())
```
MatthijsBurgh, line 19: Going from a number to a `rospy.Duration` to a number......

robot_smach_states/examples/example_hear_grasp.py
```diff
@@ -13,13 +13,13 @@ def __init__(self, robot):
...
 
     def execute(self, userdata):
-        result = self.robot.ears.recognize(spec='Bring me a <drink>',
-                                           choices={'drink': ['coke', 'fanta']},
-                                           time_out=rospy.Duration(60))
-        choices = result.choices['drink'] if result else None
-        if choices:
-            rospy.loginfo('I heared:' + repr(choices))
-        userdata.heared = choices
+        answer = self.robot.hmi.query(description='can i have a coke?',
```
MatthijsBurgh, line 16: Now we can only hear coke....

MatthijsBurgh (changes requested):
The examples should be a kind of a test, but this isn't really testing any smach state. It is only using the initialize. I would suggest to remove it.

robot_smach_states/examples/example_hear_grasp.py
shanki98: This is an example for query I suppose, so we are not hearing anything from the Robot.

MatthijsBurgh: Please restore old coke/fanta option

robot_smach_states/examples/example_hear_grasp.py
```diff
@@ -35,9 +35,10 @@ def execute(self, userdata):
 
         rospy.loginfo('I\'m going to search for {}'.format(type))
         # call robot.ed.get_closest_entity()
-        entity = {'id': 'coke-1'}
+        entity = self.robot.ed.get_closest_entity(etype=userdata.heard)
+        #entity = {'id': 'coke-1'}
```
MatthijsBurgh, line 39: Remove commented code

robot_smach_states/examples/example_hear_grasp.py
```diff
@@ -35,9 +35,10 @@ def execute(self, userdata):
...
 
-        userdata.entity = entity
+        userdata.entity = entsity
```
MatthijsBurgh, line 41: Typo

MatthijsBurgh (changes requested):

shanki98:
The example is now fixed to hear both coke and fanta. But the description is just a query and not what the robot hears. If we want the example to hear a sentence then we must use speech recognition as well in this example and this would complicate the example further. SO this example tests 
1. hmi.query
2. get_closest_entity 
3. Grammar parser

robot_smach_states/examples/example_hear_grasp.py
```diff
@@ -9,33 +9,32 @@
...
-        choices = result.choices['drink'] if result else None
-        if choices:
-            rospy.loginfo('I heared:' + repr(choices))
-        userdata.heared = choices
+        answer = self.robot.hmi.query(description='can i have a fanta?',
+                                      grammar="T -> DET coke | DET fanta \n DET -> a | an ",
+                                      target='T',
+                                      timeout=60)
+        if answer:
+            rospy.loginfo('I heard:' + answer.sentence)
```
PetervDooren, line 21: As this is an example: meant to showcase the skill. I would make the robot say this out loud. Now you can only see this in a terminal.

robot_smach_states/examples/example_hear_grasp.py
```diff
@@ -9,33 +9,32 @@
...
-        type = userdata.type
+        type = userdata.heard
 
         rospy.loginfo('I\'m going to search for {}'.format(type))
-        # call robot.ed.get_closest_entity()
-        entity = {'id': 'coke-1'}
+        entity = self.robot.ed.get_closest_entity(etype=userdata.heard)
 
         userdata.entity = entity
         return 'succeeded'
```
PetervDooren, line 40: As Matthijs also mentioned, this information is not used. But that was already an issue with this example. I don't think you need to fix this right now.

PetervDooren (dismissed):
Some suggestions for improvement, But this is alright

shanki98:
@MatthijsBurgh can this example be merged?

MatthijsBurgh (dismissed):
Merge this damn thing.

PR #1164, opened 2022-02-22
ActiveGraspDetector

P-ict0:
First version of grasp detection using the grippers position.

Created program to get the gripper's position and made new smach state for the detection.


hero_skills/src/hero_skills/hero.py
```diff
@@ -33,7 +33,8 @@ def __init__(self, wait_services=False):
         hero_arm.add_part('force_sensor', force_sensor.ForceSensor(self.robot_name, self.tf_buffer, "/" + self.robot_name + "/wrist_wrench/raw"))
         hero_arm.add_part('gripper', gripper.ParrallelGripper(self.robot_name, self.tf_buffer, 'gripper'))
         hero_arm.add_part('handover_detector', handover_detector.HandoverDetector(self.robot_name, self.tf_buffer, 'center'))
-
+	
+	hero_arm.add_part('active_grasp_detector',grasp_position_detector.GraspPositionDetector(self.robot_name, self.tf_buffer, "/" + self.robot_name + "/wrist_wrench/raw"))
```
P-ict0, line 37: Im not sure which topic I have to use here. I have not checked this

robot_smach_states/src/robot_smach_states/manipulation/active_grasp_detector.py
```diff
@@ -0,0 +1,53 @@
...
+        
+        :param arm_designator: designator that resolves to arm to check
+        :param robot: Robot to execute the state with
+        :param timeout: float amount of time the procedure may take
+        """
+        smach.State.__init__(self, outcome=['succeeded', 'failed', 'timeout'])
+        self.robot = robot
+        
+        self.arm_designator = arm_designator
+
```
P-ict0, line 26: Values for the detection have to be tested to see if there are better candidates

robot_skills/src/robot_skills/arm/grasp_position_detector.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class GraspPositionDetector(RobotPart):
+    def __init__(self, robot_name, tf_buffer, wrench_topic):
+        """
+        Class for detecting whether or not the robot is holding something
+
+        :param robot_name: Name of the robot
+        :param tf_buffer: tf2_ros.Buffer for use in RobotPart
+        :param wrench_topic: Topic to use for measurement
```
PetervDooren, line 15: wrench was used in the previous grasp detector. This one is more of a joint topic

robot_skills/src/robot_skills/arm/grasp_position_detector.py
```diff
@@ -0,0 +1,53 @@
...
+    def _joint_callback(self, msg):
+        """
+        Will be executed every time a new message is received.
+
+        :param msg: input joint message
+        """
+        if self.store_position:
+            if msg.name == 'hand_motor_joint':
+                self.current_position = msg.position
+                self.store_position = False
```
PetervDooren, line 34: This will need some work as I dont believe this is the right topic and fields.

hero_skills/src/hero_skills/hero.py
```diff
@@ -26,14 +27,20 @@ def __init__(self, wait_services=False):
...
+        hero_arm.add_part('force_sensor', force_sensor.ForceSensor(self.robot_name, self.tf_buffer,
+                                                                   "/" + self.robot_name + "/wrist_wrench/raw"))
         hero_arm.add_part('gripper', gripper.ParrallelGripper(self.robot_name, self.tf_buffer, 'gripper'))
-        hero_arm.add_part('handover_detector', handover_detector.HandoverDetector(self.robot_name, self.tf_buffer, 'center'))
+        hero_arm.add_part('handover_detector',
+                          handover_detector.HandoverDetector(self.robot_name, self.tf_buffer, 'center'))
 
+        hero_arm.add_part('active_grasp_detector',
+                          grasp_position_detector.GraspPositionDetector(self.robot_name, self.tf_buffer,
+                                                                        "/" + self.robot_name + "/wrist_wrench/raw"))
```
PetervDooren, line 43: #TODO find the right topic

robot_smach_states/src/robot_smach_states/manipulation/active_grasp_detector.py
```diff
@@ -0,0 +1,55 @@
...
+        if not arm:
+            rospy.logerr("Could not resolve arm")
+            return 'failed'
+
+        self.first_position = arm._arm.grasp_position_detector.detect()
+
+        if self.first_position is None:
+            return 'failed'
+        else:
+            arm.gripper.send_goal('close', max_torque=0.015)
```
PetervDooren, line 45: You would want max torque to be dependent on the previously used value. That might be difficult to achieve like this. At least make the torque configurable.

robot_smach_states/src/robot_smach_states/manipulation/active_grasp_detector.py
```diff
@@ -0,0 +1,55 @@
...
+
+            self.second_position = arm._arm.grasp_position_detector.detect()
```
PetervDooren, line 47: Something to verify is whether or not closing the gripper is blocking. Otherwise you may need to wait a little before taking the measurement

robot_smach_states/src/robot_smach_states/manipulation/active_grasp_detector.py
```diff
@@ -0,0 +1,55 @@
...
+        smach.State.__init__(self, outcome=['succeeded', 'failed', 'timeout'])
+        self.robot = robot
+
+        self.arm_designator = arm_designator
+
+        self.threshold_difference = 0.075  # Difference between base and final position
+        self.minimum_position = 0.2  # To be determined by testing
+
+        self.first_position = None  # First position of the gripper
+        self.second_position = None  # Position of the gripper after measuring
```
PetervDooren, line 30: These can be local variables in your execute method.

PetervDooren (changes requested):
For now the main thing is finding the correct topic on the robot to get the joint state and determine how to actually test this.

robot_skills/src/robot_skills/arm/gripper_position_detector.py
```diff
@@ -0,0 +1,52 @@
...
+import rospy
+from sensor_msgs.msg import JointState
+
+from robot_skills.robot_part import RobotPart
+
+
+class GripperPositionDetector(RobotPart):
+    def __init__(self, robot_name, tf_buffer, joint_topic):
+        """
+        Class for detecting whether the robot is holding something
```
P-ict0, line 11: This is innacurate still

robot_smach_states/src/robot_smach_states/manipulation/active_grasp_detector.py
```diff
@@ -0,0 +1,67 @@
...
+from robot_skills.arm.arms import GripperTypes
+from robot_skills.arm.gripper_position_detector import GripperPositionDetector
+
+
+class ActiveGraspDetector(smach.State):
+    REQUIRED_ARM_PROPERTIES = {"required_gripper_types": [GripperTypes.GRASPING], }
+
+    def __init__(self, robot, arm_designator, threshold_difference=0.075, minimum_position=-0.82, max_torque=0.15):
+        """
+        State for detecting whether the robot is holding something
```
P-ict0, line 17: More explanation needed

MatthijsBurgh:
Next time, better to use lower case branch names

MatthijsBurgh (commented):
Maybe add some type hinting

robot_smach_states/src/robot_smach_states/manipulation/active_grasp_detector.py
```diff
@@ -0,0 +1,76 @@
...
+                
+                if abs(first_position - second_position) < self.threshold_difference:
+                    rospy.loginfo("First position is {}\n"
+                                  "Second position is {}".format(first_position, second_position))
+                    return 'true'
+                
+                else:
+                    rospy.loginfo("First position is {}\n"
+                                  "Second position is {}".format(first_position, second_position))
+                    return 'false'
```
PetervDooren, line 72:
```suggestion
                rospy.loginfo("First position is {}\n"
                                  "Second position is {}".format(first_position, second_position))
                return abs(first_position - second_position) < self.threshold_difference
```

The logging is not necessary if everything is going right. you might want to change it to a debug statement.

robot_smach_states/src/robot_smach_states/manipulation/active_grasp_detector.py
```diff
@@ -0,0 +1,76 @@
...
+        if first_position is None:
+            rospy.loginfo("Cannot retrieve first position")
+            return 'failed'
+        elif first_position < self.minimum_position:
+            rospy.loginfo("First position is {}".format(first_position))
+            return 'Cannot determine'
+
+        else:
+            if not arm.gripper.send_goal('close', max_torque=self.max_torque):
+                rospy.logerr("Error while closing the gripper")  # Ignores the error but notifies it
```
PetervDooren, line 58: Can your method still provide a meaningful measurement even if the gripper cant close?

robot_smach_states/src/robot_smach_states/manipulation/active_grasp_detector.py
```diff
@@ -0,0 +1,76 @@
...
+            rospy.logerr("Could not resolve arm")
+            return 'failed'
+
+        first_position = arm._arm.gripper_position_detector.detect()  # First position of the gripper
+
+        if first_position is None:
+            rospy.loginfo("Cannot retrieve first position")
+            return 'failed'
+        elif first_position < self.minimum_position:
+            rospy.loginfo("First position is {}".format(first_position))
```
PetervDooren, line 53: I don't think this info is needed. You might want to change it to a debug statement.

PetervDooren (changes requested):
Think about when your method can give a reasonable result. 

Also consider which loginfos we will need if we are not actively debugging. Often we only want to log when unexpected things happen. But that is lesser problem.

robot_smach_states/src/robot_smach_states/manipulation/active_grasp_detector.py
P-ict0: Yes, that was added for testing since those values cannot be accessed from the example. Can these be removed then?

robot_smach_states/src/robot_smach_states/manipulation/active_grasp_detector.py
P-ict0: The outcome becomes "failed" if any measurement returns none

robot_smach_states/src/robot_smach_states/manipulation/active_grasp_detector.py
P-ict0: Also, the return statement wouldnt it return a bool instead of a string? If that is the case, can I change the outcome from 'true' to the boolean true? (since the return and the outcomes must match)

PetervDooren (approved):

PR #1165, opened 2022-03-02
Connection timeout

MatthijsBurgh:


PetervDooren (approved):
I am a little conflicted on this. On the one hand we provide a timeout that is essentially meaningless as it just continues as normal if the timeout is passed. So Why wait at all? 

Why not let a user decide whether or not they want to wait for the services they need?

But I realise that is not the purpose of this PR. You added a timeout parameter which was previously hardcoded. fair enough

MatthijsBurgh:
When not waiting for a service, you don't have any guarantee on the service being available when you call it. So waiting for them is not meaningless.

PR #1168, opened 2022-04-12
remove leading and trailing quotes

PetervDooren:
Task specification was a string with value "action_recipe"
Should be: action_recipe
Difficult to write down. It had ascii double quotes at the start and end of the string

### DIFF
--- a/challenge_demo/scripts/demo.py
+++ b/challenge_demo/scripts/demo.py
@@ -156,6 +156,8 @@ def main():
 
         # Dump the output json object to a string
         task_specification = json.dumps(semantics)
+        task_specification = task_specification[1:-1] # remove leading and trailing quotes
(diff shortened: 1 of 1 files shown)


challenge_demo/scripts/demo.py
```diff
@@ -156,6 +156,8 @@ def main():
 
         # Dump the output json object to a string
         task_specification = json.dumps(semantics)
+        task_specification = task_specification[1:-1] # remove leading and trailing quotes
```
MatthijsBurgh, line 159: Is this a string? Then do `task_specification.strip()`

MatthijsBurgh (changes requested):

MatthijsBurgh (approved):

PR #1170, opened 2022-04-25
Fix(Place)

PetervDooren:
Still requires testing. Have not been able to place an object yet.


robot_skills/src/robot_skills/robot.py
```diff
@@ -65,7 +65,7 @@ def __init__(self, robot_name="", tf_buffer=None, connection_timeout=DEFAULT_CON
 
         # Miscellaneous
         self.pub_target = rospy.Publisher("/target_location", geometry_msgs.msg.Pose2D, queue_size=10)
-        self.base_link_frame = self.robot_name+"/base_link"
+        self.base_link_frame = "base_link"
```
MatthijsBurgh, line 68: This is robot_specific. tf_prefix has been dropped from TF2, but this means that the urdf can't have any namespace. So `amigo/base_link` is still very valid.

So this should be configurable.

MatthijsBurgh (changes requested):

PetervDooren:
Tested on robot, robot was able to place a fanta on the table

PetervDooren:
Ping @MatthijsBurgh

MatthijsBurgh (approved):

PR #1189, opened 2022-06-07
(smach_states) check more laser beams while waiting for door

MatthijsBurgh:


PetervDooren:
What problem is this solving?

MatthijsBurgh:
Because of laser filters, values can be NaN. When in the range there is only NaNs. It assumes the door is not opened. So you need to take more beams into account. Though you don't want to take too much, so we limit it to max 21.

PetervDooren:
When would points be NaN? I cant imagine the robot standing in front of the door and not getting laser points

MatthijsBurgh:
When the laser filter decides it needs to be none. So that really depends on the situation. When the door is closed. The laser filter will probably not do that much. But when it opens, it can see some ghost points, depending on what is in the arena.

And we are not talking about getting good laser point  at all. We are talking about good laser points right in front of the robot.

PetervDooren:
Did this ever happen during testing? because this seems very rare.

MatthijsBurgh:
> Did this ever happen during testing? because this seems very rare.

Yes, it did. We are not that bored...

MatthijsBurgh:
@PetervDooren still has been merged in #1226, except for the remaining comments. Which is a good addition IMO.

PR #1194, opened 2022-06-18
fix(find_people_in_room): Cast filter to list

reinzor:


reinzor:
Python2 --> Python3 leftover

PR #1200, opened 2022-06-23
Receptionist fix drink designators

PetervDooren:
Fix for the robot saying 'you are paul and you like [None]'
this was caused because `guest_drink_des` was redefined locally rather than using the input. 

I made it so that we pass the drink as a string variable rather than a HMI result as I think this makes processing the information less comples. After all, it does not matter how we know the persons favourite drink

This change was tested on the robot and the speech regarding the drink works as it should now. (the robot still cant hear very well)

### DIFF
--- a/challenge_receptionist/src/challenge_receptionist/learn_guest.py
+++ b/challenge_receptionist/src/challenge_receptionist/learn_guest.py
@@ -8,6 +8,7 @@ from robot_smach_states.reset import ResetArms
 
 import robot_smach_states.util.designators as ds
 import smach
+from smach import cb_interface, CBState
 from robocup_knowledge import load_knowledge
 from hmi import HMIResult
 
@@ -30,7 +31,7 @@ class LearnGuest(smach.StateMachine):
         smach.StateMachine.__init__(self, outcomes=['succeeded', 'failed', 'aborted'])
 
         self.drink_spec_des = ds.Designator(challenge_knowledge.common.drink_spec, name='drink_spec')
-        guest_drink_des = ds.VariableDesignator(resolve_type=HMIResult, name='guest_drink')
+        guest_drink_hmi_des = ds.VariableDesignator(resolve_type=HMIResult, name='guest_drink')
         guest_drink_name_des = ds.FieldOfHMIResult(guest_drink_des, semantics_path='drink', name='guest_drinkname')
 
         with self:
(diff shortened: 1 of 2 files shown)


challenge_receptionist/src/challenge_receptionist/learn_guest.py
```diff
@@ -101,14 +102,25 @@ def __init__(self, robot,
...
+            @cb_interface(outcomes=['done'])
+            def _write_drink_name(ud, answer, drink_name_designator):
+                name = answer.resolve()
+                drink_name_designator.write(name)
+                return 'done'
+
+            smach.StateMachine.add("WRITE_DRINK_NAME",
+                                   CBState(_write_drink_name,
+                                           cb_args=[guest_drink_name_des, guest_drink_des.writeable]),
+                                   transitions={'done': 'SAY_DRINK_CORRECT_NAME'})
```
MatthijsBurgh, line 119: What is happening here? We get the value of one designator and write it to another one?

MatthijsBurgh (changes requested):

challenge_receptionist/src/challenge_receptionist/learn_guest.py
```diff
@@ -30,8 +31,8 @@ def __init__(self, robot,
         smach.StateMachine.__init__(self, outcomes=['succeeded', 'failed', 'aborted'])
 
         self.drink_spec_des = ds.Designator(challenge_knowledge.common.drink_spec, name='drink_spec')
-        guest_drink_des = ds.VariableDesignator(resolve_type=HMIResult, name='guest_drink')
-        guest_drink_name_des = ds.FieldOfHMIResult(guest_drink_des, semantics_path='drink', name='guest_drinkname')
+        guest_drink_hmi_des = ds.VariableDesignator(resolve_type=HMIResult, name='guest_drink')
+        guest_drink_name_des = ds.FieldOfHMIResult(guest_drink_hmi_des, semantics_path='drink', name='guest_drinkname')
```
MatthijsBurgh, line 35: This should get you the info from the HMIResult...

MatthijsBurgh (changes requested):

MatthijsBurgh:
Redefining input vars is so wrong. So keeping that while writing some extra code to workaround this mess is not good.

MatthijsBurgh:
I recommend to go for alternative PR #1201

LarsJanssenTUe:
See https://github.com/tue-robotics/tue_robocup/pull/1201

PR #1225, opened 2022-08-30
check goal not state

PetervDooren:
This was a bug. Previous implementation did not give the intended behaviour

### DIFF
--- a/robot_skills/src/robot_skills/arm/gripper.py
+++ b/robot_skills/src/robot_skills/arm/gripper.py
@@ -106,7 +106,7 @@ class ParrallelGripper(Gripper):
         if goal.command.direction == GripperState.OPEN:
             if self.occupied_by is not None:
                 rospy.logerr("ParrallelGripper.send_goal open is called but there is still an entity with id '%s' \
-                occupying the gripper, removing this entity. Please update your worldmodel" % self.occupied_by.uuid)
+                occupying the gripper, removing this entity from the robot interface." % self.occupied_by.uuid)
             self.occupied_by = None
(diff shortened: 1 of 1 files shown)


robot_skills/src/robot_skills/arm/gripper.py
```diff
@@ -103,10 +103,10 @@ def send_goal(self, state, timeout=5.0, max_torque=0.1): #Todo: should send goal
 
         self._ac_gripper.send_goal(goal)
 
-        if state == GripperState.OPEN:
+        if goal.command.direction == GripperState.OPEN:
             if self.occupied_by is not None:
                 rospy.logerr("ParrallelGripper.send_goal open is called but there is still an entity with id '%s' \
-                occupying the gripper, please update the world model and remove this entity" % self.occupied_by.uuid)
+                occupying the gripper, removing this entity. Please update your worldmodel" % self.occupied_by.uuid)
```
MatthijsBurgh, line 109: I think the meaning was that it should be removed from the world model. So in that case the new log is incorrect.

(Also both changes should not be in 1 commit)

MatthijsBurgh (changes requested):

robot_skills/src/robot_skills/arm/gripper.py
PetervDooren: That does not at all reflect what happens here or what was done before. The entity held by the robot is not present in the worldmodel. So that old log was definitely incorrect.

MatthijsBurgh (approved):

PR #1228, opened 2022-09-06
Rwc2022 challenge take out the garbage

PetervDooren:


challenge_take_out_the_garbage/src/challenge_take_out_the_garbage/startup_inside_arena.py
```diff
@@ -0,0 +1,44 @@
+from __future__ import absolute_import
```
PetervDooren, line 1: If we can make this generic lets do it

challenge_take_out_the_garbage/src/challenge_take_out_the_garbage/startup_inside_arena.py
```diff
@@ -0,0 +1,44 @@
...
+
+# ROS
+import smach
+
+# TU/e Robotics
+from robot_skills.robot import Robot
+from robot_smach_states import check_ebutton, utility
+
+
+class StartChallengeRobustInsideArena(smach.StateMachine):
```
MatthijsBurgh, line 11: This is not needed

MatthijsBurgh (changes requested):

challenge_take_out_the_garbage/src/challenge_take_out_the_garbage/pick_up.py
```diff
@@ -261,6 +286,9 @@ def __init__(self, robot, trashbin_designator, arm_designator):
         smach.StateMachine.__init__(self, outcomes=["succeeded", "failed", "aborted"])
         place_pose_designator = EmptySpotDesignator(robot, trashbin_designator, arm_designator)
 
+        if not place_pose_designator:
```
MatthijsBurgh, line 289: This is not possible. Designator doesn't have a `__bool__`. This will just check if the object is not None or similar.

MatthijsBurgh (changes requested):

PR #1234, opened 2022-09-06
Rwc2022 challenge set the table

PetervDooren:


MatthijsBurgh (approved):
Updating these images it not good for the repo size

PR #1236, opened 2022-09-06
Rwc2022 challenge where is this

PetervDooren:


GustavoDCC:
From running, hero-challenge-where-it-is, I got this error:
```
[ERROR][/state_machine][1669755424.266936]: waypoint_constraint function: No entity could be resolved from designator 'EntityByIdDesignator(uuid=where_is_this_information_point, name=None)'
[ERROR][/state_machine][1669755424.269136]: function: <function NavigateToWaypoint.__init__.<locals>.<lambda> at 0x7f8bec152550> in combine_constraints resulted in None
[ERROR][/state_machine][1669755424.271981]: combine_constraints function did not resolve to any constraints! functions combined were [<function NavigateToWaypoint.__init__.<locals>.<lambda> at 0x7f8bec152550>]
```
@MatthijsBurgh, I add what you suggested:  tf_buffer.transform(...) has a zero timeout, which makes it not robust. It needs to have a little timeout. A PR should be open about this, opened by Janno.

PetervDooren:
Since this is very stale I will close this PR. I also suggest to delete the branch but wont do so yet in case we want to keep it for some reason.

PR #1242, opened 2022-09-16
(robot_smach_states) small improvements in test

MatthijsBurgh:


PetervDooren:
Does a missing main mean these tests were previously not preformed at all?

MatthijsBurgh:
> Does a missing main mean these tests were previously not preformed at all?

No, but now you can them separately.

PetervDooren (approved):

PR #1244, opened 2022-09-24
(test_tools) record video from camera

PetervDooren:
copied from rwc2022 hero_bringup.
This would be a better place for it.

### DIFF
--- a/test_tools/scripts/record_video_from_camera.py
+++ b/test_tools/scripts/record_video_from_camera.py
@@ -1,5 +1,6 @@
 #! /usr/bin/env python3
 
+import argparse
 import sys
 import rospy
 import cv2
@@ -41,12 +42,17 @@ class ImageConverter:
 
 
 def main():
-    image_topic = "/hero/head_rgbd_sensor/rgb/image_raw"
-    video_name = "video.avi"
-    if len(sys.argv) > 1:
-        image_topic = sys.argv[1]
-        if len(sys.argv) > 2:
-            video_name = sys.argv[2]
+    parser = argparse.ArgumentParser(
+        description='Record video from a camera image topic',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument("--topic", default="/hero/head_rgbd_sensor/rgb/image_raw",
+                        help="image topic of type sensor_msgs/Image encoded in bgr8")
+    parser.add_argument("--file", default="video.avi",
+                        help="name of the video file to save to")
+    args = parser.parse_args()
+
+    image_topic = args.topic
+    video_name = args.file
 
     rospy.init_node("save_video", anonymous=True)
     rospy.loginfo(f"Starting video recorder for topic {image_topic} saving to video file {video_name}")
--- a/test_tools/scripts/record_video_from_camera.py
+++ b/test_tools/scripts/record_video_from_camera.py
@@ -1,0 +1,48 @@
+#! /usr/bin/env python3
+
+import rospy
+import cv2
+from cv_bridge import CvBridge, CvBridgeError
+from sensor_msgs.msg import Image
+
+
+class image_converter(object):
+    def __init__(self, topic_name, file_name, fps=20):
+        self.fps = fps
+        self.file_name = file_name
+        self.bridge = CvBridge()
+        self.image_sub = rospy.Subscriber(topic_name, Image, self.callback)
+        self.video_writer = None
+
+    def callback(self, data):
+        try:
+            cv_image = self.bridge.imgmsg_to_cv2(data, "bgr8")
+        except CvBridgeError as e:
+            rospy.logerr(e)
+            raise
+        else:
+            if self.video_writer is None:
+                rows, cols, _ = cv_image.shape
+                fourcc = cv2.VideoWriter_fourcc(*"MJPG")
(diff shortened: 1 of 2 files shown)


test_tools/scripts/record_video_from_camera.py
```diff
@@ -0,0 +1,48 @@
+#! /usr/bin/env python3
+
+import rospy
+import cv2
+from cv_bridge import CvBridge, CvBridgeError
+from sensor_msgs.msg import Image
+
+
+class image_converter(object):
```
MatthijsBurgh, line 9:
```suggestion
class ImageConverter:
```

test_tools/scripts/record_video_from_camera.py
```diff
@@ -0,0 +1,48 @@
...
+            self.video_writer.release()
+        rospy.loginfo(f"Saving video file {self.file_name}")
+
+    def __del__(self):
+        self.clean_shutdown()
+
+
+def main():
+    rospy.init_node("save_video", anonymous=True)
+    ic = image_converter("/hero/head_rgbd_sensor/rgb/image_raw", "video.avi")
```
MatthijsBurgh, line 42:
```suggestion
    ic = ImageConverter("image_raw", "video.avi")
```

MatthijsBurgh (changes requested):
This does require cv_bridge and opencv to be added as dependency to test_tools

test_tools/scripts/record_video_from_camera.py
```diff
@@ -31,15 +32,25 @@ def callback(self, data):
...
+            rospy.loginfo(f"Saving video file {self.file_name}")
+        else:
+            rospy.loginfo(f"No video saved, no image received on topic")
 
     def __del__(self):
         self.clean_shutdown()
 
 
 def main():
+    image_topic = "/hero/head_rgbd_sensor/rgb/image_raw"
```
MatthijsBurgh, line 44: I think remapping might be better for this.

test_tools/scripts/record_video_from_camera.py
```diff
@@ -31,15 +32,25 @@ def callback(self, data):
...
+    video_name = "video.avi"
+    if len(sys.argv) > 1:
+        image_topic = sys.argv[1]
+        if len(sys.argv) > 2:
+            video_name = sys.argv[2]
```
MatthijsBurgh, line 49: This is not robust for remapping or other ROS args.  Use
```py
myargv = rospy.myargv()
```

P.S. Ideally you use `argparse` for the parsing instead of this positional behaviour.

MatthijsBurgh (changes requested):

test_tools/scripts/record_video_from_camera.py
PetervDooren: ease of use is more important to me in this case. you can still specify another topic if needed.

test_tools/scripts/record_video_from_camera.py
PetervDooren: This code will only be run separately. it was never meant to be used with other ros tags

test_tools/scripts/record_video_from_camera.py
MatthijsBurgh: Making it too easy too use, will make people lazy. And people will not understand anymore what is happening under the hood. Keep that in mind.

test_tools/scripts/record_video_from_camera.py
MatthijsBurgh: No, but someday some idiot will do it.

test_tools/scripts/record_video_from_camera.py
```diff
@@ -0,0 +1,59 @@
...
+        self.clean_shutdown()
+
+
+def main():
+    image_topic = "/hero/head_rgbd_sensor/rgb/image_raw"
+    video_name = "video.avi"
+    if len(sys.argv) > 1:
+        image_topic = sys.argv[1]
+        if len(sys.argv) > 2:
+            video_name = sys.argv[2]
```
MatthijsBurgh, line 49:
```suggestion
    myargv = rospy.myargv()
    if len(myargv) > 1:
        image_topic = myargv[1]
        if len(myargv) > 2:
            video_name = myargv[2]
```

@PetervDooren luckily I didn't became lazy :p

test_tools/scripts/record_video_from_camera.py
PetervDooren: There is no need to know what is happening under the hood. As long as the documentation is up to date. (I added an argparser to improve on the previous version for this.)

test_tools/scripts/record_video_from_camera.py
```diff
@@ -0,0 +1,65 @@
...
+
+def main():
+    parser = argparse.ArgumentParser(
+        description='Record video from a camera image topic',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument("--topic", default="/hero/head_rgbd_sensor/rgb/image_raw",
+                        help="image topic of type sensor_msgs/Image encoded in bgr8")
+    parser.add_argument("--file", default="video.avi",
+                        help="name of the video file to save to")
+    args = parser.parse_args()
```
MatthijsBurgh, line 52:
```suggestion
    args = parser.parse_args(rospy.myargv())
```

MatthijsBurgh (approved):

PR #1251, opened 2022-10-21
Fix/grasping benchmark

PetervDooren:
tested the grasping benchmark. Which was broken. 
Fixed proper handling of assertionerrors
usage of framestamped for saving data
and used waypoints that exist in the impuls environment. (note it is not stated that the benchmark test needs to be performed in this environment specifically in the docs, imo this should be done in a readme/wiki on how to run the benchmark test)

### DIFF
--- a/robot_smach_states/benchmark/benchmark_grasping.py
+++ b/robot_smach_states/benchmark/benchmark_grasping.py
@@ -70,8 +70,7 @@ def single_item(robot, results_writer, cls, support, waypoint, inspect_from_area
               'timestamp': datetime.datetime.now().strftime('%Y-%m-%d %H:%M:%S')}
 
     try:
-        say_announce = Say(robot, sentence="Please put a {cls} {search_area} the {support}"
-                           .format(cls=cls,support=support,search_area=search_area))
+        say_announce = Say(robot, sentence=f"Please put a {cls} {search_area} the {support}")
         say_announce.execute()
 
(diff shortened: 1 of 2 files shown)


robot_smach_states/benchmark/benchmark_grasping.py
```diff
@@ -111,10 +111,10 @@ def single_item(robot, results_writer, cls, support, waypoint, inspect_from_area
 
                 entity = grasp_entity.resolve()  # type: Entity
                 if entity:
-                    vector_stamped = entity.pose  # type: VectorStamped
-                    record['x'] = '{:.3f}'.format(vector_stamped.vector.x())
-                    record['y'] = '{:.3f}'.format(vector_stamped.vector.y())
-                    record['z'] = '{:.3f}'.format(vector_stamped.vector.z())
+                    frame_stamped = entity.pose  # type: VectorStamped
```
MatthijsBurgh, line 114: Change the typehinting

MatthijsBurgh (changes requested):
One small change

robot_smach_states/benchmark/benchmark_grasping.py
PetervDooren:
```suggestion
                    frame_stamped = entity.pose  # type: FrameStamped
```

MatthijsBurgh (approved):

PR #1258, opened 2022-11-15
People recognition

jlunenburg:


robot_skills/src/robot_skills/perception.py
```diff
@@ -184,11 +189,83 @@ def detect_faces(self, image=None, stamp=False):
         """
         if not image:
             image = self.get_image()
+        # ToDo: create a consistent API: either return a tuple, a message or a list
```
PetervDooren, line 192: could you create an issue for this?

robot_skills/src/robot_skills/perception.py
```diff
@@ -184,11 +189,83 @@ def detect_faces(self, image=None, stamp=False):
...
+        expected_operator_pos: typing.Optional[kdl.Vector] = None,
+        threshold: float = 0.5,
+    ) -> Recognition:
+        """
+        For all provided recognitions, project the ROI and determine the one closest to the expected operator position,
+        computed on the floor plane.
+
+        :param recognitions: recognitions
+        :param expected_operator_pos: expected position of the operator w.r.t. the robot
+        :param threshold: if the distance between the recognition and the expected pos exceed the threshold, an error
```
PetervDooren, line 235:
```suggestion
        :param threshold: if the distance between the recognition and the expected pos exceed the threshold, a Runtime error
```

robot_skills/src/robot_skills/perception.py
```diff
@@ -184,11 +189,83 @@ def detect_faces(self, image=None, stamp=False):
...
+        is raised
+        :return: Recognition closed to the expected position
+        :raises: RuntimeError
+        """
+        frame_id = f"{self.robot_name}/base_link"
```
PetervDooren, line 240: The robot object can have different base link names. This is stored as robot.base_link_frame or something like that. Perhaps this should be added to the constructor of the perception part and stored as an attribute

robot_skills/src/robot_skills/perception.py
```diff
@@ -184,11 +189,83 @@ def detect_faces(self, image=None, stamp=False):
...
+        projected_recognitions = [(rcg, self.project_roi(rcg.roi, frame_id)) for rcg in recognitions]
+        expected_operator_pos = kdl.Vector(1.0, 0.0, 0.0) if expected_operator_pos is None else expected_operator_pos
```
PetervDooren, line 242: You could use the 1,0,0 vector as default instead of None in that case. That would also make it clear in the functions signature what the default behaviour is.

robot_smach_states/examples/human_interaction/example_people_recognition.py
```diff
@@ -0,0 +1,85 @@
...
+        Example statemachine to test people learning and recognition
+
+        :param robot: robot API object
+        """
+        smach.StateMachine.__init__(self, outcomes=["Done"])
+
+        name_designator = VariableDesignator(resolve_type=str)
+
+        with self:
+            smach.StateMachine.add(
```
PetervDooren, line 29: It might be lipservice but I think it might be handy for the robot to say "I will now learn." since the operator will be away from the laptop they might not be able to monitor the terminal

robot_smach_states/src/robot_smach_states/human_interaction/recognize_person.py
```diff
@@ -0,0 +1,79 @@
...
+        raise RuntimeError("Recognition does not contain probabilities")
+    best_match = max(distribution.probabilities, key=lambda cp: cp.probability)  # type: CategoryProbability
+    if best_match.probability < threshold:
+        raise RuntimeError(
+            f"Best match has a probability of {best_match.probability} which is lower than the threshold {threshold}"
+        )
+    return best_match.label
+
+
+class RecognizePerson(smach.State):
```
PetervDooren, line 37: can we make the distinction between getting no detections i.e. not seeing anyone and seeing someone but not recognising them? Would that be something desirable?

PetervDooren (changes requested):

robot_skills/src/robot_skills/perception.py
jlunenburg: Done, see #1264

robot_skills/src/robot_skills/perception.py
jlunenburg: A `kdl.Vector` is mutable. Hence, though that suggestion would work in this specific case, I don' t consider it good programming practice. Consider the following example:
```
In [1]: import PyKDL as kdl                                                                          

In [2]: def foo(bar = kdl.Vector()): 
   ...:     print(f"\nBefore: {bar}") 
   ...:     bar.x(bar.x() + 1.0) 
   ...:     print(f"After: {bar}") 
   ...:                                                                                              

In [3]: foo()                                                                                        

Before: [           0,           0,           0]
After: [           1,           0,           0]

In [4]: foo()                                                                                        

Before: [           1,           0,           0]
After: [           2,           0,           0]
```

robot_skills/src/robot_skills/perception.py
jlunenburg: See 8904256a000980011173ba30f016b191dc1f8a79

robot_smach_states/examples/human_interaction/example_people_recognition.py
jlunenburg: This was already present in the `LearnPerson` state (see [human_interaction.py](https://github.com/tue-robotics/tue_robocup/blob/master/robot_smach_states/src/robot_smach_states/human_interaction/human_interaction.py). Therefore I considered it overkill

robot_smach_states/src/robot_smach_states/human_interaction/recognize_person.py
jlunenburg: Perhaps, I added the distinction: 2f723d8ac56b36b342aabb56cb9459e0ff16ebc0

PetervDooren (approved):

PR #1263, opened 2022-12-17
Add Show Map from final

MatthijsBurgh:
Best to review commits separately.

### DIFF
--- a/robot_smach_states/src/robot_smach_states/human_interaction/show_image.py
+++ b/robot_smach_states/src/robot_smach_states/human_interaction/show_image.py
@@ -1,30 +1,33 @@
-from __future__ import absolute_import
-
-# ROS
-import os
 import rospy
 import smach
 
+from robot_smach_states.util.designators import check_type
 
-class ShowImageState(smach.State):
-    """
-    This state allows images to be shown on the hero display given the package in which the image is located and the
-    path of the image within that package
-    """
 
-    def __init__(self, robot, image_filename, seconds=5.0):
+class ShowImage(smach.State):
+    """
+    This state allows images to be shown on the robot display
+    """
+    def __init__(self, robot, filename, duration=5.0):
         """
+        Constructor
+
         :param robot: the robot object
-        :param image_filename: string describing the path to the image (absolute path incl. file extension)
+        :param filename: string or designator describing the path to the image (absolute path incl. file extension)
+        :param duration: Duration for which the image should be shown in seconds
         """
-        super(ShowImageState, self).__init__(outcomes=['succeeded', 'failed'])
-        self._filename = image_filename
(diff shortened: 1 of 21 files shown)


robot_smach_states/src/robot_smach_states/human_interaction/show_image.py
```diff
@@ -1,30 +1,30 @@
...
 import rospy
 import smach
 
+from robot_smach_states.util.designators import check_type
 
-class ShowImageState(smach.State):
-    """
-    This state allows images to be shown on the hero display given the package in which the image is located and the
-    path of the image within that package
-    """
```
PetervDooren, line 6: Do keep the high level description of what the state does

PetervDooren (changes requested):

robot_smach_states/src/robot_smach_states/world_model/get_show_map.py
```diff
@@ -0,0 +1,110 @@
...
+    from robot_skills import get_robot
+
+    rospy.init_node(os.path.splitext("test_" + os.path.basename(__file__))[0])
+    r = get_robot("hero")
+    # Works for impuls environment
+    entities_des = VariableDesignator(["dining_room"], resolve_type=[str])
+    mark_ids_des = VariableDesignator(["initial_pose", "salon_table"], resolve_type=[str])
+    plan_points_des = VariableDesignator(resolve_type=Vector)
+    sm = GetMapAndShow(r, entities_des, mark_ids_des, plan_points_des)
+    sm.execute()
```
PetervDooren, line 110: Should this be here because it is very specific? Or do we want to keep this seperatly as an example?

PetervDooren:
Please check why CI is failing

robot_smach_states/src/robot_smach_states/human_interaction/show_image.py
MatthijsBurgh: Oh that wasn't the plan

MatthijsBurgh:
> Please check why CI is failing

Yes, I made a mistake. I will fix it.

robot_smach_states/src/robot_smach_states/human_interaction/show_image.py
MatthijsBurgh: I have put it back, but I have changed it too.

robot_smach_states/src/robot_smach_states/world_model/get_show_map.py
MatthijsBurgh: I will move this to examples. Did we ever decide that we want all `__main__` removed from the library? If we did, we should write it down. So we don't forget it again.

PetervDooren (approved):

PR #1272, opened 2023-03-04
Cleanup of grab.py

P-ict0:
After various tests of grabbing objects from different surfaces we can safely remove the "lifting" part (`robot_smach_states/manipulation/grab.py`) after it grasped the object and go directly to "retract". Doesn't necessarily improve speed but simplifies the code.

Also added some type hinting.

### DIFF
--- a/robot_smach_states/src/robot_smach_states/manipulation/grab.py
+++ b/robot_smach_states/src/robot_smach_states/manipulation/grab.py
@@ -79,7 +79,7 @@ class PickUp(smach.State):
                                "required_goals": ["carrying_pose"], }
 
     def __init__(self, robot: Robot, arm: ArmDesignator, grab_entity: Designator,
-                 check_occupancy: bool = False, visual_servoing: bool = False) -> None:
+                 check_occupancy: bool = False) -> None:
         """
         Pick up an item given an arm and an entity to be picked up
 
@@ -87,7 +87,6 @@ class PickUp(smach.State):
         :param arm: Designator that resolves to the arm to grasp with
         :param grab_entity: Designator that resolves to the entity to grab. e.g EntityByIdDesignator
         :param check_occupancy: Indicates whether to check if gripper is occupied
-        :param visual_servoing: Indicates whether to use visual servoing
         """
         smach.State.__init__(self, outcomes=['succeeded', 'failed'])
 
@@ -98,7 +97,6 @@ class PickUp(smach.State):
         self.grab_entity_designator = grab_entity
         self._gpd = GraspPointDeterminant(robot)
         self._check_occupancy = check_occupancy
-        self._use_visual_servoing = visual_servoing
 
         assert self.robot.get_arm(**self.REQUIRED_ARM_PROPERTIES) is not None,\
             "None of the available arms meets all this class's requirements: {}".format(self.REQUIRED_ARM_PROPERTIES)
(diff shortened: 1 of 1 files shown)


robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -197,27 +198,15 @@ def execute(self, userdata=None):
...
-            rospy.logerr('Failed lift')
-
         # Retract
         goal_bl = self.robot.tf_buffer.transform(grasp_framestamped, self.robot.base_link_frame)
         rospy.loginfo('Start retracting')
         roll = 0.0
 
         goal_bl.frame.p.x(goal_bl.frame.p.x() - 0.1)  # Retract 10 cm
-        goal_bl.frame.p.z(goal_bl.frame.p.z() + 0.05)  # Go 5 cm higher
+        goal_bl.frame.p.z(goal_bl.frame.p.z() + 0.01)  # Go 10 cm higher
```
PetervDooren, line 209: This is 1 cm higher. Did you mean
```suggestion
        goal_bl.frame.p.z(goal_bl.frame.p.z() + 0.1)  # Go 10 cm higher
```

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -176,15 +178,14 @@ def execute(self, userdata=None):
...
-        #     arm.gripper.send_goal('close', timeout=None)
-        #     return 'failed'
+        if self._use_visual_servoing:
+            rospy.loginfo('Starting Pre-grasp')
+            if not arm.send_goal(goal_bl.x, goal_bl.y, goal_bl.z, 0, 0, 0, frame_id=self.robot.base_link_frame,
+                                 timeout=20, pre_grasp=True, first_joint_pos_only=True):
+                rospy.logerr('Pre-grasp failed:')
+                arm.reset()
+                arm.gripper.send_goal('close', timeout=None)
+                return 'failed'
```
PetervDooren, line 188: This appears to do exactly what is being done below here. Which is move the arm to the goal pose. Visual servoing does not appear to be happening here.
Question 1: How should the behaviour of the robot change with this flag.
Question 2: was this change tested.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -80,13 +78,16 @@ class PickUp(smach.State):
     REQUIRED_ARM_PROPERTIES = {"required_gripper_types": [GripperTypes.GRASPING],
                                "required_goals": ["carrying_pose"], }
 
-    def __init__(self, robot, arm, grab_entity, check_occupancy=False):
+    def __init__(self, robot: Robot, arm: ArmDesignator, grab_entity: Designator,
+                 check_occupancy: bool = False, visual_servoing: bool = False) -> None:
```
PetervDooren, line 82: At some point we need to think about how to manage these auxilary functions. We cant keep adding flags to turn stuff on or off.
But for now this is alright. In the future we may need to think about adding a pre-hook and a post-hook to be more general.

PetervDooren (changes requested):

robot_smach_states/src/robot_smach_states/manipulation/grab.py
P-ict0: This was first commented out, but indeed if it does exactly what we do below we can remove it since it was not used before aswell.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
P-ict0: If we end up removing the visual servoing check below in lines 181-188, we can delete the flag

PetervDooren (approved):

PetervDooren:
Hi @P-ict0, don't forget to re-request a review. That way it shows up in my list of reviews requested. I just ran into this one by accident this time.

PR #1275, opened 2023-03-28
[place] Removed aggressive force drive after placing object

P-ict0:
Hero force drived backwards after it has placed an object

### DIFF
--- a/robot_smach_states/src/robot_smach_states/manipulation/place.py
+++ b/robot_smach_states/src/robot_smach_states/manipulation/place.py
@@ -158,7 +158,9 @@ class Put(smach.State):
                                                 frame_id=self._robot.base_link_frame),
                       timeout=0.0)
 
-        self._robot.base.force_drive(-0.125, 0, 0, 1.5)
+        # Wait for arm to finish their motions
+        arm.wait_for_motion_done()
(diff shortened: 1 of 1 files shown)


robot_smach_states/src/robot_smach_states/manipulation/place.py
```diff
@@ -158,7 +158,9 @@ def execute(self, userdata=None):
                                                 frame_id=self._robot.base_link_frame),
                       timeout=0.0)
 
-        self._robot.base.force_drive(-0.125, 0, 0, 1.5)
+        # Wait for arm to finish their motions
+        arm.wait_for_motion_done()
+        self._robot.base.force_drive(-0.125, 0, 0, 2, ax=0.5)
```
MatthijsBurgh, line 163: Why do we drive further? I think the reduction by the constrained acceleration is very limited.

P-ict0: We can keep it at 1.5 as it was before, indeed the acceleration doesn't make much of an impact there

MatthijsBurgh: Yes make it 1,5 again. Increasing the loop rate might also smooth the driving more.

PetervDooren (approved):

PR #1294, opened 2023-04-23
PicoVoice robotpart and state

MatthijsBurgh:


robot_skills/src/robot_skills/picovoice.py
```diff
@@ -0,0 +1,40 @@
...
+        super().__init__(robot_name=robot_name, tf_buffer=tf_buffer)
+        self._action_client = self.create_simple_action_client(f"/{robot_name}/get_intent", GetIntentAction)
+
+    def get_intent(self, context_url: str, require_endpoint=True, timeout: float = 10.0) -> HMIResult:
+        """
+        get_intent
+
+        :param context_url:
+        :param require_endpoint:
+        :param timeout:
```
PetervDooren, line 20: Please add a bit of context what these do. Not all users will be familiar with picovoice workings.

robot_skills/src/robot_skills/picovoice.py
```diff
@@ -0,0 +1,40 @@
...
+        if not self._action_client.send_goal_and_wait(goal, execute_timeout=timeout, preempt_timeout=timeout):
+            rospy.logerr("PicoVoice.get_intent: failed")
+            raise TimeoutException("PicoVoice.get_intent: failed")
+
+        result = self._action_client.get_result()  # type: GetIntentResult
+        rospy.logdebug(f"PicoVoice.get_intent: {result=}")
+
+        if result is None:
+            rospy.logerr("PicoVoice.get_intent: result: None")
+            raise TimeoutException("PicoVoice.get_intent: result: None")
```
PetervDooren, line 36: These are not strictly timeoutexceptions. They also conflict with the docstring which states that it will only raise a timeoutexception in case of timeout

robot_skills/src/robot_skills/picovoice.py
```diff
@@ -0,0 +1,40 @@
...
+        if not result.is_understood:
+            rospy.logwarn("PicoVoice.get_intent: Not understood")
+            raise TimeoutException("PicoVoice.get_intent: Not understood")
```
PetervDooren, line 39: These are not strictly timeoutexceptions. They also conflict with the docstring which states that it will only raise a timeoutexception in case of timeout

PetervDooren (changes requested):
Minor changes requested

MatthijsBurgh:
Merged in https://github.com/tue-robotics/tue_robocup/pull/1310

PR #1316, opened 2023-05-16
Receptionist find empty seat2

MatthijsBurgh:


challenge_receptionist/src/challenge_receptionist/find_empty_seat2.py
```diff
@@ -0,0 +1,177 @@
...
+        return "SeatsInRoomDesignator({}, {}, {}, {})".format(self.robot, self.seat_ids, self.room, self.name)
+
+
+class FindEmptySeat(smach.StateMachine):
+    """
+    Iterate over all seat-type objects and check that their 'on-top-of' volume is empty
+    That can be done with an Inspect and then query for any Entities inside that volume.
+    If there are none, then the seat is empty
+    """
+
```
PetervDooren, line 56: Add information on the arguments in the docstring. What do the arguments mean?

challenge_receptionist/src/challenge_receptionist/find_empty_seat2.py
```diff
@@ -0,0 +1,177 @@
+#! /usr/bin/env python
```
PetervDooren, line 1: This file is not needed unless you want to test this version and the original side by side. I think it was originally created as a tournament hack. But now you may feel free to make changes to find_empty_seat.py

PetervDooren (commented):
No real functional changes here. Added some general tips but I dont think these changes add value just yet.

PR #1317, opened 2023-05-29
(Place) Removed useless motion of arm and improved docs

P-ict0:
Current implementation: Robot goes to the `prepare_place` pose BEFORE navigating to the place pose.
PR: Robot now doesn't move to the `prepare_place` pose anymore at any moment.

Also changed some asserts to use more correct type checking

### DIFF
--- a/robot_smach_states/src/robot_smach_states/manipulation/place.py
+++ b/robot_smach_states/src/robot_smach_states/manipulation/place.py
@@ -11,7 +11,7 @@ from robot_skills.arm.arms import PublicArm, GripperTypes
 from .place_designator import EmptySpotDesignator
 from ..navigation.navigate_to_place import NavigateToPlace
 from ..utility import LockDesignator, ResolveArm, check_arm_requirements
-from ..util.designators import check_type
+from ..util.designators import check_type, check_resolve_type
 from ..util.designators.utility import LockingDesignator
 from robot_smach_states.world_model.world_model import Inspect
 
@@ -165,8 +165,8 @@ class Place(smach.StateMachine):
         smach.StateMachine.__init__(self, outcomes=['done', 'failed'])
 
         # Check types or designator resolve types
-        assert(item_to_place.resolve_type == Entity or type(item_to_place) == Entity)
-        assert(arm.resolve_type == PublicArm or type(arm) == PublicArm)
+        assert(check_resolve_type(item_to_place, Entity) or isinstance(item_to_place, Entity))
+        assert(check_resolve_type(arm, PublicArm) or isinstance(arm, PublicArm))
 
         # parse place volume
         if place_volume is not None:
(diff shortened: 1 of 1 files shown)


PetervDooren (approved):
Depending on the implementation used we may need to put prepare place where it is now commented. It would be nice to figure that out fast.

PetervDooren:
It is already broken for robots other than Hero. We don't have the infrastructure to properly support 'any' robot.
What happened here is that the functionality of PUT was implicitly changed due to switching to collision avoidance. As long as the other robots keep a similar functionality it will remain consistent.

PetervDooren (dismissed):

robot_smach_states/src/robot_smach_states/manipulation/place.py
```diff
@@ -167,7 +167,6 @@ def __init__(self, robot, item_to_place, place_pose, arm, place_volume=None, upd
         # Check types or designator resolve types
         assert(item_to_place.resolve_type == Entity or type(item_to_place) == Entity)
         assert(arm.resolve_type == PublicArm or type(arm) == PublicArm)
```
MatthijsBurgh, line 169: Should use `check_resolve_type`

robot_smach_states/src/robot_smach_states/manipulation/place.py
```diff
@@ -165,8 +165,8 @@ def __init__(self, robot, item_to_place, place_pose, arm, place_volume=None, upd
         smach.StateMachine.__init__(self, outcomes=['done', 'failed'])
 
         # Check types or designator resolve types
-        assert(item_to_place.resolve_type == Entity or type(item_to_place) == Entity)
-        assert(arm.resolve_type == PublicArm or type(arm) == PublicArm)
```
MatthijsBurgh, line 205: Sorry, use `check_type`

MatthijsBurgh (changes requested):

PetervDooren:
@MatthijsBurgh, the check_type is a correct improvement but it has nothing to do with the changes in this PR. The changes here are fine and can be merged without worrying about changing check type as well. 
If we want to change this check type I suggest doing it in a small PR rather than delaying this one any further.

robot_smach_states/src/robot_smach_states/manipulation/place.py
```diff
@@ -201,9 +165,8 @@ def __init__(self, robot, item_to_place, place_pose, arm, place_volume=None, upd
...
-        #assert(place_volume.resolve_type == str or (type(place_volume) == str))
+        assert(check_type(item_to_place, Entity) or isinstance(item_to_place, Entity))
+        assert(check_type(arm, PublicArm) or isinstance(arm, PublicArm))
```
MatthijsBurgh, line 169:
```suggestion
        check_type(item_to_place, Entity)
        check_type(arm, PublicArm)
```

MatthijsBurgh:
@PetervDooren changes were already made to these lines. So therefore the changes need to be correct and complete.

I applied the changes.

PetervDooren:
> @PetervDooren changes were already made to these lines. So therefore the changes need to be correct and complete.
> 
> I applied the changes.

No they were not. A commented line was removed. That does not make Rodrigo responsible for fixing all the lines near that. That is not how this works.

PetervDooren (approved):

PR #1319, opened 2023-06-05
(place.py) Typing and formatting

P-ict0:
Removed unused variable and added type-hinting

### DIFF
--- a/robot_smach_states/src/robot_smach_states/manipulation/place.py
+++ b/robot_smach_states/src/robot_smach_states/manipulation/place.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import, print_function
+from typing import Union, Optional
 
 # ROS
 from pykdl_ros import FrameStamped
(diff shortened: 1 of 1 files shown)


PetervDooren (approved):

robot_smach_states/src/robot_smach_states/manipulation/place.py
```diff
@@ -8,18 +9,20 @@
...
 from ..util.designators.utility import LockingDesignator
 from robot_smach_states.world_model.world_model import Inspect
+from ..util.designators.core import Designator
 
 
 class Put(smach.State):
     REQUIRED_ARM_PROPERTIES = {"required_gripper_types": [GripperTypes.GRASPING], }
 
-    def __init__(self, robot, item_to_place, placement_pose, arm):
+    def __init__(self, robot: Robot, item_to_place: Entity, placement_pose: FrameStamped, arm: PublicArm) -> None:
```
MatthijsBurgh, line 25: Arm is a designator....

P-ict0: All of the other variables are designators then according to the docs, should we change them too then?

MatthijsBurgh: Yes, Designators can be typed. So that should be done. The `check_type` in the constructor should be replaced by `check_resolve_type`. As only designators are allowed.

PR #1323, opened 2023-07-15
[rwc2023][remainder]

MatthijsBurgh:


robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -206,7 +206,7 @@ def execute(self, userdata=None) -> str:
         if not arm.send_goal(goal_bl, timeout=0.0, allowed_touch_objects=[grab_entity.uuid]):
             rospy.logerr('Failed retract')
         arm.wait_for_motion_done()
-        self.robot.base.force_drive(-0.125, 0, 0, 2.0)
+        self.robot.base.force_drive(-0.2, 0, 0, 2.0)
```
PetervDooren, line 209: This is a bigger change in behaviour. I would reverse this and we can tune it again if it becomes a problem.

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -217,11 +217,11 @@ def execute(self, userdata=None) -> str:
         # Remove pose from ED as we are holding the object in the gripper
         self.robot.ed.update_entity(uuid=grab_entity.uuid, remove_pose=True)
 
-        arm.wait_for_motion_done(cancel=True)
-
```
PetervDooren, line 221: While it often wont change the behaviour, I do think its important to keep this check

robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -217,11 +217,11 @@ def execute(self, userdata=None) -> str:
...
         # Carrying pose
         arm.send_joint_goal('carrying_pose', timeout=0.0)
 
+        arm.wait_for_motion_done(cancel=True)
+
```
PetervDooren, line 224: The robot can start driving while going to the carrying pose. Often this goes well and it makes the robot faster. I prefer the old way of doing this.

robot_smach_states/src/robot_smach_states/manipulation/place.py
```diff
@@ -221,8 +224,14 @@ def __init__(self, robot: Robot, item_to_place: Designator[Entity], place_pose:
...
-                                   transitions={'succeeded': 'done',
-                                                'failed': 'failed'})
+                                   transitions={'succeeded': 'UNLOCK_DESIGNATOR_SUCCESS',
+                                                'failed': 'UNLOCK_DESIGNATOR_FAILED'})
+
+            smach.StateMachine.add('UNLOCK_DESIGNATOR_SUCCESS', UnlockDesignator(locking_place_designator, False),
+                                   transitions={'unlocked': 'done'})
+
+            smach.StateMachine.add('UNLOCK_DESIGNATOR_FAILED', UnlockDesignator(locking_place_designator, False),
+                                   transitions={'unlocked': 'failed'})
```
PetervDooren, line 234: Why does the designator need to be unlocked? If another place is needed wouldnt you need to make a new designator?

robot_smach_states/src/robot_smach_states/manipulation/place_designator.py
```diff
@@ -114,10 +117,13 @@ def _resolve(self):
         rospy.logerr("EmptySpotDesignator: Could not find an empty spot")
         return None
 
-    def _is_poi_unoccupied(self, frame_stamped, surface_entity):
+    def _is_poi_unoccupied(self, frame_stamped: FrameStamped, surface_entity: Entity, volume_id: Optional[str] = None):
```
PetervDooren, line 120: this function is starting to become a mess. But I think that is unavoidable in its current format. In the future we should think about moving this to ED and its plugins.

robot_smach_states/src/robot_smach_states/navigation/follow_operator.py
```diff
@@ -175,7 +175,7 @@ def _register_operator(self):
         """
         start_time = rospy.Time.now()
 
-        self._robot.head.look_at_standing_person(distance=1.5)
+        self._robot.head.look_at_standing_person(distance=1.5, height=1.3)
```
PetervDooren, line 178: too many magic numbers. This would be alright in a challenge but in a smach state its going to give us headaches later on.

robot_smach_states/src/robot_smach_states/navigation/follow_operator.py
```diff
@@ -218,7 +218,7 @@ def _register_operator(self):
                                                          block=True)
                                 self._robot.speech.speak("Just in case...",
                                                          block=False)
-                                self._robot.head.look_at_standing_person()
+                                self._robot.head.look_at_standing_person(distance=1.5, height=1.3)
```
PetervDooren, line 221: see above

robot_smach_states/src/robot_smach_states/navigation/follow_operator.py
```diff
@@ -487,18 +506,19 @@ def _recover_operator(self, look_distance: float = 2.0):
         if not self._learn_face:
             return False
         rospy.loginfo("Trying to recover the operator")
-        self._robot.head.look_at_standing_person()
+        self._robot.head.look_at_standing_person(distance=1.5, height=1.3)
```
PetervDooren, line 509: see above

robot_smach_states/src/robot_smach_states/utility.py
```diff
@@ -368,25 +368,29 @@ def execute(self, userdata=None):
 
 
 class LockDesignator(smach.State):
-    def __init__(self, locking_designator):
+    def __init__(self, locking_designator, resolve_print=True):
```
PetervDooren, line 371: I dont like this as an argument. Like this you get functions with 10+ flag arguments to trigger different behaviours. And this one just changes wether something gets printed.
I say just print it every time.

PetervDooren (changes requested):

robot_smach_states/src/robot_smach_states/utility.py
MatthijsBurgh: The problem with the print is that it causes the designator to be resolved. While you only want the designator to be resolved when you use it. This isn't a problem for a normal designator, but is is for a locking designator.

robot_smach_states/src/robot_smach_states/navigation/follow_operator.py
MatthijsBurgh: The question is how to split the functionality...

PR #1326, opened 2023-07-25
Rwc2023 challenge serve breakfast

PetervDooren:
issue #1325 keeps track of the issues that are still in the code. I suggest to first merge the rwc2023 version and then make smaller PRs to fix these smaller issues.

Issue #1325, opened 2023-07-25
Challenge serve breakfast


PetervDooren:
As of 
#1326  challenge serve breakfast still has several issues. 

wishlist for features

- [ ] Get configuration from robocup knowledge. Now the pick and place locations are still hardcoded.
- [ ] grasp detection does not work yet. It should be implemented but when a grasp is failed the robot still proceeds as normal: fixed by #1327 
- [ ] hardcoded navigation poses w.r.t the table. this means if you change table the robot will drive into.

These are the features that can be fixed by good bookkeeping and configuration.

PR #1327, opened 2023-07-25
Use handover detector to detect failed handover

PetervDooren:
Merge after #1326 
Fixes one of the issues in #1325

### DIFF
--- a/challenge_serve_breakfast/src/challenge_serve_breakfast/navigate_to_and_pick_item.py
+++ b/challenge_serve_breakfast/src/challenge_serve_breakfast/navigate_to_and_pick_item.py
@@ -112,6 +112,9 @@ class PickItem(StateMachine):
                                   'failed': "failed",
                                   'cannot_determine': "SAY_SOMETHING_WENT_WRONG"}
                      )
+            self.add("SAY_SOMETHING_WENT_WRONG", Say(robot, "Oops, it seems I missed it. Lets try again"),
+                     transitions={"spoken": "ASK_USER2"})
+            self.add("ASK_USER2", CBState(_ask_user), transitions={"succeeded": "CARRYING_POSE", "failed": "failed"})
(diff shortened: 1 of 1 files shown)


PetervDooren (commented):
Tested on the robot. Worked like a charm

MatthijsBurgh (changes requested):
I think the issue was always it requires tuning per item it is holding. I think that downside is still there. So works like a charm for a subset of items probably. Please prove me wrong.

PetervDooren:
That subset is all objects that are not paper thin. Which includes at least all objects used in serve breakfast. I tested it with a spoon and bowl

MatthijsBurgh (changes requested):
Add comment about the used tuning for detector works for what set of items, i.e. their dimensions

PetervDooren:
There was no tuning done in this PR. it just uses the grasp detector. Grasp detector is tuned for how accurate it can detect things. No need to add a comment here.

MatthijsBurgh:
There is no comment either in the challenge or the ActiveGraspDetector about which the dimensions is works on. It is only mentioned when the object is too thin/small it doesn't work. No clue about what that threshold is.\

PetervDooren:
Docstring of ActiveGraspDetector states anything wider than the threshold (0.075 rad on the gripper) can be detected. You can say this is not clear enough but that is not an issue for this PR
        
       State for detecting whether the robot is holding something using the gripper position.

        Stores current position of the hand motor joint, slightly closes the gripper and then compares the new
        position with the first one. If the difference is bigger than the threshold, robot is holding something

        If the object is too small/thin it will not be able to determine

        :param robot: Robot to execute the state with
        :param arm_designator: designator that resolves to arm to check
        :param threshold_difference: Difference between base and final position
        :param minimum_position: Minimum position to assume that the gripper is holding something
        :param max_torque: Max torque of the gripper to perform the test with

MatthijsBurgh (approved):

PR #1335, opened 2023-10-31
Use grasp detector in grab.py

PetervDooren:
implemented and tested on the robot. This also required re-tuning the grasp detector. It is now a bit conservative and it will not be able to detect objects of less than centimeter across.

### DIFF
--- a/robot_smach_states/src/robot_smach_states/manipulation/grab.py
+++ b/robot_smach_states/src/robot_smach_states/manipulation/grab.py
@@ -14,7 +14,10 @@ from robot_skills.arm.arms import PublicArm, GripperTypes
 from ..utility import check_arm_requirements, ResolveArm
 from ..util.designators import check_type
 from ..navigation.navigate_to_grasp import NavigateToGrasp
+
+from ..manipulation.active_grasp_detector import ActiveGraspDetector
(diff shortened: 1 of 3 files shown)


robot_smach_states/src/robot_smach_states/manipulation/grab.py
```diff
@@ -339,9 +342,20 @@ def __init__(self, robot: Robot, item: Designator, arm: ArmDesignator, room: Des
...
 
+            smach.StateMachine.add("CHECK_PICK_SUCCESSFUL",
+                                  ActiveGraspDetector(robot, arm),
+                                  transitions={'true': "done",
+                                              'false': "SAY_MISSED_GRASP",
+                                              'failed': "RESET_FAILURE",
+                                              'cannot_determine': "SAY_MISSED_GRASP"}
+                                  )
+
+            smach.StateMachine.add("SAY_MISSED_GRASP", Say(robot, "Oops, it seems I missed it. Lets try again"),
```
PetervDooren, line 356:
```suggestion
            smach.StateMachine.add("SAY_MISSED_GRASP", Say(robot, "Oops, it seems I missed it."),
```

PR #1341, opened 2024-03-21
Demo peter

PetervDooren:
Some suggestions based on our demo today. Fixes the following problems:

1. reinstates wait for trigger. I find this useful for demos since the presenter can determine the tempo of the presentation. Otherwise the robot is constantly asking for validation. 
2. Fix the timeout logic for not hearing the operator. Previous implementation just continued with the loop which crashes if the sentence parameter was not initialised. Otherwise it will just redo the previous command. Now it says "oops we couldn't do it." and waits for the next trigger.

### DIFF
--- a/challenge_demo/scripts/demo.py
+++ b/challenge_demo/scripts/demo.py
@@ -123,7 +123,7 @@ def main():
                                    etype="waypoint")
 
             # Listen for the new task
-            while True:
+            for i in range(timeout_count):
                 try:
                     if is_sim_mode():
                         sentence, semantics = robot.hmi.query(description="",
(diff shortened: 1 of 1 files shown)


PetervDooren:
@MatthijsBurgh please don't add commits to this PR. If you have suggestions for the demo please make a PR to the update_demo branch which Lotte oversees. These changes have been tested on the robot.

challenge_demo/scripts/demo.py
```diff
@@ -132,14 +132,13 @@ def main():
...
-                    if timeout_count >= 3:
-                        rospy.logwarn("[GPSR] Timeout_count: {}".format(timeout_count))
-                        break
+            else:
+                rospy.logwarn("[GPSR] speech timed out {timeout_tries} times, aborting")
+                robot.speech.speak("I am sorry but I cannot hear you. Let's not try again.")
+                continue  # Continue the outer while loop
 
             # check if we have heard this correctly
             robot.speech.speak('I heard %s, is this correct?' % sentence)
```
PetervDooren, line 144: Something I noticed but did not fix. The robot will say the action it understood rather than what was said. i.e. it will say "I heard hand-over is this correct?" or "I heard navigate-to, is this correct? or "I heard say, is this correct"

MatthijsBurgh: Yes, I think this is the new behaviour. But @lottemessing should confirm this.

